PR #2, opened 2016-03-14
'1st' Commit(s)

atareshawty:
Set up structure of app, including express boilerplate


atareshawty:
Added React.js for the server side template engine

PR #4, opened 2016-03-22
Fixed readme and added navbar partial page

atareshawty:


atareshawty:
Right now, it's just the bootstrap example navbar. No idea what's going there yet

PR #9, opened 2016-03-31
Data models

atareshawty:
Starting to build data models just as a framework. Just want to see it up on heroku to make sure everything is working so far


atareshawty:
node-gyp needs g++/gcc 4.8 to compile all of the necessary c++/c stuff. This was primarily to build the dependencies for the excel and node-xlsx packages

.travis.yml
```diff
@@ -6,8 +6,19 @@ notifications: # set notification options
 sudo: false
 
 node_js:
-  - "4.1"
+  - "5.4.1"
```
btmills: > [docs](https://docs.travis-ci.com/user/languages/javascript-with-nodejs#Choosing-Node-versions-to-test-against): Choosing “5” or “4” will use the latest version available in the 5.x or 4.x releases.

Why `"5.4.1"` and not `"5"`? Is Heroku giving you an older version?

atareshawty: I was having trouble getting node-gyp builds to work, so I was trying to make Travis as cloase to my local machine as possible. It can be changed back to 5 now that I figured out the actual problem

PR #13, opened 2016-04-04
WIP Challenge Creation

atareshawty:
Successfully creating a challenge relies on a user logged in. We're waiting on OSU to get permission to use the single sign-on. Not much more can be built without it


atareshawty:
All of the controller actions won't actually work through the browser. All functionality will be tested through Jasmine which will mock requests and responses through the controllers. As soon as we have login, things will work through the browser

atareshawty:
Lots about the data models were changed, so I'm going to do one PR per model to get everything implemented

PR #17, opened 2016-04-06
WIP Nginx

atareshawty:
Nginx will give us the ability to hit osumbchallenges.dev rather than localhost:3000. The main reason for this is to be able to have a redirect URL to give to OSU's SSO for local testing (osumbchallenges.dev/login/callback or something like that)


atareshawty:
Wrong branch...

PR #18, opened 2016-04-06
Added nginx.conf.erb file to set up config for osumbchallenges.dev

atareshawty:
Nginx will give us the ability to hit osumbchallenges.dev rather than localhost:3000. The main reason for this is to be able to have a redirect URL to give to OSU's SSO for local testing (osumbchallenges.dev/login/callback or something like that)


Brewfile
```diff
@@ -1,2 +1,3 @@
 brew 'postgresql'
 brew 'node'
+brew "nginx"
```
tarebyte: @atareshawty you need to add `brew tap github/bootstrap` to the top of this file see: https://github.com/github/homebrew-bootstrap#usage

config/NGINX.md
```diff
@@ -0,0 +1,24 @@
+# If you want to use Nginx for a url without a port
```
tarebyte: I'm sorry I remember why this works for me out of the box.

I have my machine setup with https://github.com/mikemcquaid/strap which sets this up for you.

tarebyte:
Outside of my two comments this looks good @atareshawty

Brewfile
atareshawty: Why? It works without

tarebyte: Because other if someone installs this fresh the nginx part of th bootstrap won't work.

PR #21, opened 2016-04-07
Jasmine Testing

atareshawty:
Unit tests shouldn't care about authenticating users, so we have to mock stuff! So, Jasmine. The point of the PR is to set up a framework for unit testing with mocks and to give a few examples of specs


atareshawty:
Right now, we're stuck in an awkward place. Controllers should ideally be testing for user authentication, but we can't do auth yet, so they don't. Once OSU SSO is done, we'll have to update some of the mocks.

PR #25, opened 2016-04-13
Updated data models for cleaner challenge creation

atareshawty:
Upon trying to write the code to create a challenge, I found that with how the data models stood, this was really difficult. This is a change to make that process cleaner. All of the relations can be found in models/index.js.


atareshawty:
This *should* be the framework for how the data models work now. Basic routing is set up and we'll go through a model per PR to get everything implemented

PR #26, opened 2016-04-14
[WIP] Users

atareshawty:
Implementing the users model/controller/views and such (with a few other things)


atareshawty:
NOTE: we're doing all implementation through testing and we don't have authentication in place, so nothing will work in the browser

atareshawty:
@tarebyte UsersControllerSpec.js is where the magic happens

atareshawty:
There's a much better way to be doing auth that removes the need for almost all of these tests...

PR #31, opened 2016-04-30
Auth!

atareshawty:
## Wait Alex, I thought you were going to use OSU's SSO?!?!?
We were! Unfortunately, I found out today that might involve a 2 month security vetting from OSU's security counsel and I ain't about that. We're doing our own local auth.

### TODO:

- [x] Get Some Nice-ish UI login things
- [x] Add some sort of session store (Redis) as Express's Memory Store isn't for production
- [x] Add logout functionality
- [x] Have more than just a catch all unauthorized view/route for a little more explanation
- [x] Make sure things work on Heroku
- [x] Clean things up and make pretty code

## Well... How are you going to do it?
You're asking all sorts of questions aren't you?

We're using Passport.js and just a local authentication strategy with Redis for storing user sessions. The plan is to preload all members of the band in the database and give them a unique password for their first login. The username will be the name.number. Then, they can change the password if they like. 

__Note: We're going to have to build on Heroku from this branch before merging to the actual staging branch so we don't have to make a butt-ton of PR's to trigger Heroku builds__


atareshawty:
Heroku kind of works!! I don't 100% trust it yet, but we can log in and store a session (I think). I'm talking to myself 😦    `¯\_(ツ)_/¯`

atareshawty:
__To test on Heroku, the username/password is any name.number from the mock data__

btmills:
Just wanna say I really enjoyed reading this thread :fire:

PR #32, opened 2016-05-01
[WIP] Make challenge

atareshawty:
# Making Challenges
Finally! What this app is actually supposed to do 🎉  🎈 

## TODO
- [x] Solidify data models
- [x] Document the data models (to avoid future headache)
- [x] Build out some mock data
- [x] Test all of the things!
- [x] Actually implement
- [x] Make sure it works on Heroku

## How
When an eligible user makes a challenge, a row in the challenger table will be created with the proper info. If the spot is 'fully challenged', we'll flag the spot so that it's no longer shown as valid to other challengers. Only when the challenges have been approved by the director will the results table be filled in (without any result data)

## Timing is Important
Most of this app is time based 😿 ... When testing time based things, we have to be really careful. For instance, when creating mock performances, we need to provide a time window. If the current time when running the test is past the time we made the mock performance, we won't have an upcoming performance. Because of this, we only get so long to just run `jasmine` rather than `npm test`. The former just runs the test, the latter populates the database with the mock data, pushing back the performance window time. ⌚ 🕐  __Edit: running the tests on the last day of the month will break them... Time is hard__


atareshawty:
https://github.com/osumb/challenges/pull/32/commits/a941c9920b1ac29168e75baa6397c70355a58d92 For documentation

btmills:
If you need help manipulating time in a sane fashion, [Moment.js](http://momentjs.com/) is nifty.

atareshawty:
@btmills I am using Moment.js, but probably not correctly I guess.

config/config.js
```diff
@@ -15,18 +15,6 @@ module.exports = {
     'port': process.env.PORT || 3000
   },
   'test': {
-    'mockPerformance': {name: 'Bowling Green Game', openAt: moment(buildDate()).add({days: 1}).format(), closeAt: moment(buildDate()).add({days: 1, hours: 3}).format()}
+    'mockPerformance': {name: 'Bowling Green Game', openAt: moment().startOf('day').add({days: 1}).format(), closeAt: moment().startOf('day').add({days: 1, hours: 3}).format()}
```
atareshawty: @btmills Thanks lol Much better now

config/config.js
```diff
@@ -15,18 +15,6 @@ module.exports = {
...
-  let now = new Date();
-  let year = now.getFullYear(), month = now.getMonth() + 1, day = now.getUTCDate();
-  if (month < 10) {
-    month = '0' + month;
-  }
-  if (day < 10) {
-    day = '0' + day;
-  }
-  return new String(year) + '-' + month + '-' + day;
-}
```
btmills: Oh yeah. This was nasty.

atareshawty:
There are things to clean up, but this is enough for a demo for the squad leader retreat

PR #34, opened 2016-05-08
Staging

atareshawty:
# What?!? We're finally merging to master?
Yeah. I was asked to give a talk on it for the squad leader retreat (May 14th), so we should have a 'MVP' out there.

## You're so cool using words like MVP. What's it going to include?
TODO:
- [x] Fake Profile Data/Views (they don't have to be super pretty)
- [x] Challenge Select
- [x] Set up heroku for production app (it'll be exactly like staging, so that's cool)
- [x] Practice presentation cause you know you're not going to do well on the fly

__**If there's time, hopefully it can look at least semi OSU official__
__***Since the staging branch is protected, still make PRs to staging to contribute to this PR__


atareshawty:
The heroku app (osumbchallenges.herokuapp.com) now has a web hook listening for merges to the master branch. It has postgres, redis, and the session secret all ready to go as environment variables. The Procfile on master is bad, but the one in this PR will work
- Set up Heroku for production ✔️

atareshawty:
Audible! We won't need need any official domain/upper tier dyno app quite yet (don't let it linger though). The next step we need is to get the forms ready to go for a squad leader to make decisions on challenges.

Idea: We'll want a field on the results table to reference the user who signed off on the challenge and which user declared the winner/wrote comments.

atareshawty:
Still need squad leader views, but we're headed to master

PR #42, opened 2016-05-24
New models

atareshawty:
First installment of our quest to kill sequelize. We're doing straight sql and it's gonna be awesome


atareshawty:
Tests are hella broken 😿

config/config.js
```diff
@@ -20,9 +20,9 @@ module.exports = {
     'port': process.env.PORT || 3000
   },
   'test': {
-    'mockPerformance': [
-      {name: 'Bowling Green Game', openAt: moment().startOf('day').add().format(), closeAt: moment().startOf('day').add({days: 1, hours: 3}).format()},
-      {name: 'Buckeye Invite', openAt: moment().startOf('month').add().format(), closeAt: moment().startOf('month').add({days: 1, hours: 3}).format()}
+    'mockPerformances': [
```
atareshawty: I don't like that living here, but I'll punt on that for now

models/Challenge.js
```diff
@@ -0,0 +1,30 @@
...
+const utils = require('../utils');
+
+class Challenge {
+  makeChallenge(userId, spotId) {
+    const sql = queries.makeChallenge;
+
+    return new Promise((resolve, reject) => {
+      const client = utils.db.createClient();
+
+      client.connect();
```
atareshawty: A connection to the db is opened for every function call to our model classes. At a larger scale, that will probably lead to some performance issues. If it actually leads to a problem, I'll optimize

models/Result.js
```diff
@@ -1,25 +1,36 @@
 'use strict';
```
atareshawty: De we need these all over the place now?

models/queries/make-challenge-transaction.sql
```diff
@@ -0,0 +1,11 @@
+-- Things we need to check before for making a challenge
+-- 1) Check if the user hasn't already made a challenge for the current performance (and is elligible)
+-- 2) Check for availability of the spot (if spotChallenged count is less than full)
+-- If those two checks pass, we'll make the challenge
+-- 1) Create the row into the challenge table
+-- 2) Increment the spotChallenged count for the spot we're challenging
+-- 3) If the spot is full, we'll create a row in the results table
+-- We'll consider breaking this up into functions
+START TRANSACTION;
```
atareshawty: This will be fun to write

models/Challenge.js
```diff
@@ -0,0 +1,30 @@
...
+      const query = client.query(sql, [userId, spotId]);
+
+      query.on('row', (result) => {
+        client.end();
+        resolve(result);
+      });
+
+      query.on('error', (err) => {
+        client.end();
+        reject({ error: err });
```
atareshawty: In some places, we're rejecting with `reject({ error: err })` in others we're just calling `reject(err)`. Pick one and be consistent

PR #43, opened 2016-05-25
Fix controllers

atareshawty:
All of the things are broken!


controllers/Users.js
```diff
@@ -21,45 +19,21 @@ function UsersController() {
...
-    promise.then((challengeableUsers) => {
-      res.render('challengeSelect', {user: req.user, challengeableUsers: challengeableUsers});
-    });
-
-    promise.catch(() => {
-      res.render('error');
-    });
-
-    return promise;
+    Promise.all([Challenge.getAllChallengeablePeopleForUser(req.user), Performance.getNext()])
```
atareshawty: I doubt this needs to be two separate queries

models/Challenge.js
```diff
@@ -2,6 +2,52 @@ const queries = require('./queries');
 const utils = require('../utils');
 
 class Challenge {
+  getAllChallengeablePeopleForUser(user) {
+    const sql = 'SELECT * FROM spots AS S, users AS U WHERE u.instrument = $1 AND u.part = $2 AND u.eligible = FALSE AND S.id = U.spotId';
```
atareshawty: Maybe include performance here (or another query) so we don't have to make two queries

atareshawty:
There might be places in this code where a model promise might not be resolving. For instance
```js
query.on('row', (result) => {
   client.end();
   resolve(result);
});
query.on('end', () => client.end());
```
If the query doesn't return any rows, the promise never gets resolved and we hang for ever 😮

controllers/Users.js
atareshawty: This definitely doesn't need to be two separate queries

models/User.js
```diff
@@ -1,26 +1,18 @@
...
     return new Promise((resolve, reject) => {
       const client = utils.db.createClient();
-      const queryString = 'SELECT * FROM "Users" WHERE "nameNumber" = $1;';
+      const queryString = 'SELECT * FROM users WHERE nameNumber = $1;';
 
       client.connect();
       client.on('error', (err) => {reject(err);});
 
       const query = client.query(queryString, [nameNumber]);
       query.on('row', (result) => {
```
atareshawty: Add this. If we don't find a user, the promise won't ever resolve
```js
query.on('end', () => {
  client.end();
  resolve(null);
}
```

models/schema.sql
```diff
@@ -109,6 +109,7 @@ CREATE TABLE users (
   eligible boolean NOT NULL DEFAULT FALSE,
 	squadLeader boolean NOT NULL DEFAULT FALSE,
 	admin boolean NOT NULL DEFAULT FALSE,
+	alternate boolean NOT NULL DEFAULT FALSE,
```
atareshawty: @btmills I suddenly feel the need to alphabetize everything

btmills: You shall be assimilated.

controllers/Challenges.js
```diff
@@ -0,0 +1,21 @@
...
+  this.new = (req, res) => {
+    const userId = req.user.nameNumber,
+      spotId = req.body['challenge-form'],
+      performanceId = req.params.performanceId;
+
+    Challenge.makeChallenge(userId, spotId, performanceId)
+      .then(() => {
+        req.user.eligible = false;
+        console.log(`${req.user.name} successfully challenged for ${spotId}`); //TODO: logging
+        res.render('challengeSuccess', { user: req.user, spotId });
```
atareshawty: Would we want to do a redirect to profile with a flash message?

models/Challenge.js
```diff
@@ -1,29 +1,94 @@
-const queries = require('./queries');
 const utils = require('../utils');
 
 class Challenge {
-  makeChallenge(userId, spotId) {
-    const sql = queries.makeChallenge;
+  getAllChallengeablePeopleForUser(user) {
+    const sql = 'SELECT * FROM spots AS S, users AS U WHERE u.instrument = $1 AND u.part = $2 AND u.eligible = FALSE AND S.id = U.spotId';
```
atareshawty:
```sql
ORDER BY S.id
```

atareshawty: Then we're going to want to check if the spot has been challenged and add some styling it it is

models/Challenge.js
```diff
@@ -1,29 +1,94 @@
...
   }
+
+  parseChallenge(challenge) {
+    return {
+      spotId: challenge.spotid,
+      performanceName: challenge.name
+    };
+  }
+
+  parseChallengeAblePerson(challenge) {
```
atareshawty: Let's check if the spot has been challenged and add some styling if it is

PR #49, opened 2016-06-01
Kill sequelize

atareshawty:
Let's test the crap out of this on staging first


atareshawty:
Down with sequelize!! :shipit:

PR #50, opened 2016-06-01
Why Heroku no work?

atareshawty:
Ugh

### DIFF
--- a/app.js
+++ b/app.js
@@ -32,7 +32,7 @@ app.use(session( {
   secret: process.env.PASSPORT_SECRET || 'notMuchOfASecret',
   resave: true,
   saveUninitialized: true,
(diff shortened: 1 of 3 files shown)


app.js
```diff
@@ -32,7 +32,7 @@ app.use(session( {
   secret: process.env.PASSPORT_SECRET || 'notMuchOfASecret',
   resave: true,
   saveUninitialized: true,
-  session: redis ? new RedisStore({client: redis}) : undefined
+  store: redis ? new RedisStore({ client: redis }) : null
```
atareshawty: Store, not session... Always read the docs, kids

atareshawty:
This PR served two purposes. One to fix whatever the hell was wrong with my db setup script. The second was to fix our session store issue on Heroku and run in strict mode so we get `const` and other things.

PR #59, opened 2016-06-02
Webpack scss

atareshawty:
And it actually works on Heroku!


atareshawty:
What we're doing here is pretty simple, so there probably won't be a need for different webpack environment configs

PR #79, opened 2016-06-30
Email

atareshawty:
Right now the email is getting sent to my address (for testing) when a challenge is successful. It worked on staging, so we'll test on prod and then I'll merge


atareshawty:
It works on prod 🚢

PR #82, opened 2016-07-07
File organization

atareshawty:
Yaaasss. More organized routes -> controllers -> models/views thanks @tarebyte


atareshawty:
Works on staging

PR #83, opened 2016-07-11
Switch spots

atareshawty:
### In This PR
--------------------
- [ ] When a result is approved, the results controller checks if all results have been approved

If all results for the current performance are approved, the spots are switched. We have to do a special sorting so spots don't get messed up. The sorting is
- Regular vs Alternate
- Alternate vs Alternate
- Regular vs Regular

__Results with only one participant are done separately__

This has been tested against the current mock data with some additional results to add complexity.
This process runs under the assumption that the last result is approved during the time frame for the performance

### DIFF
--- a/controllers/Results.js
+++ b/controllers/Results.js
@@ -9,11 +9,10 @@ function ResultsController() {
     .then(() => {
       const { id } = req.session.currentPerformance;
 
-      res.json({ message: 'success!' });
+      res.json({ success: true });
       Result.checkAllDoneForPerformance(id)
       .then(done => {
         if (done) {
-          console.log('Switching spots for performance id', id);
           Result.switchSpotsForPerformance(id);
         }
       })
@@ -21,7 +20,7 @@ function ResultsController() {
     })
     .catch((err) => {
       console.error(err);
-      res.json({ err });
+      res.json({ success: false });
     });
   };
 
--- a/controllers/Results.js
+++ b/controllers/Results.js
@@ -6,7 +6,19 @@ function ResultsController() {
     const { ids } = req.body;
 
     Result.approve(ids)
-    .then(() => res.json({ message: 'success!' }))
+    .then(() => {
+      const { id } = req.session.currentPerformance;
+
+      res.json({ message: 'success!' });
+      Result.checkAllDoneForPerformance(id)
+      .then(done => {
+        if (done) {
+          console.log('Switching spots for performance id', id);
+          Result.switchSpotsForPerformance(id);
+        }
+      })
+      .catch(err => console.error(err));
+    })
     .catch((err) => {
       console.error(err);
       res.json({ err });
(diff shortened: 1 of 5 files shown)


db/schema.sql
```diff
@@ -68,16 +158,7 @@ BEGIN
...
-  -- There are two ways we could grab the other user. If they're also challenging the spot
-  -- or if the spot isn't open, we need the current user associated with the spot
-  IF (spotOpen AND cCount + 1 = 2) THEN
-    INSERT INTO results (performanceId, spotId, firstNameNumber, secondNameNumber, pending)
-    VALUES (pId, sId, (SELECT userNameNumber FROM challenges WHERE performanceId = pId AND spotId = sId), uId, TRUE);
-  ELSIF (NOT spotOpen AND cCount + 1 = 1) THEN
-    INSERT INTO results (performanceId, spotId, firstNameNumber, secondNameNumber, pending, needsApproval)
-    VALUES (pId, sId, uId, (SELECT nameNumber FROM users WHERE spotId = sId), TRUE, FALSE);
-  END IF;
+
```
atareshawty: Results won't be created until the challenge window cron job runs. Seems simpler that way

controllers/Results.js
```diff
@@ -6,7 +6,19 @@ function ResultsController() {
     const { ids } = req.body;
 
     Result.approve(ids)
-    .then(() => res.json({ message: 'success!' }))
+    .then(() => {
+      const { id } = req.session.currentPerformance;
+
+      res.json({ message: 'success!' });
```
atareshawty:
```js
res.json({ success: true });
```

controllers/Results.js
```diff
@@ -6,7 +6,19 @@ function ResultsController() {
...
+        if (done) {
+          console.log('Switching spots for performance id', id);
+          Result.switchSpotsForPerformance(id);
+        }
+      })
+      .catch(err => console.error(err));
+    })
     .catch((err) => {
       console.error(err);
       res.json({ err });
```
atareshawty:
```js
res.json({ success: false });
```

models/Result.js
```diff
@@ -139,6 +189,50 @@ module.exports = class Results {
...
+        });
+      });
+
+      switchQuery.on('error', err => {
+        client.end();
+        console.error(err);
+      });
+    });
+
+    resultsQuery.on('error', err => console.error(err));
```
atareshawty:
```js
resultsQuery.on('error', err => {
  console.error(err);
  client.end();
});
```

atareshawty:
Everything checks out on staging

db/schema.sql
```diff
@@ -31,12 +31,102 @@ RETURNS void AS $$
...
+      IF userOne = winner AND spotOne <> winnerSpot THEN
+        UPDATE users SET spotId = winnerSpot WHERE nameNumber = userOne;
+        UPDATE users SET spotId = spotOne WHERE nameNumber = userTwo;
+      ELSIF winner = userTwo AND spotTwo <> winnerSpot THEN
+        UPDATE users SET spotId = winnerSpot WHERE nameNumber = userTwo;
+        UPDATE users SET spotId = spotTwo WHERE nameNumber = userOne;
+      END IF;
+    END IF;
+
+    -- If two regulars were involved in the challenge
```
atareshawty: This comment is wrong...

Issue #78, opened 2016-06-30
TODO


atareshawty:
These are the things that for sure need to get done to be usable:

- [x] Email user when he/she signs up for challenge
- [x] Spot switch when challenge result is confirmed
- [x] Challenge list exporting when signup time is over ([Node-Schedule](https://github.com/node-schedule/node-schedule) when a performance is created)
- [x] Logging on challenge success (for my sanity and debugging)
- [x] #87
- [x] Admin things that would be useful to have (probably need to meet with Tess)
- [x] #114 
- [x] #110
- [x] Just learned today that different directors are going to be responsible for approving different instrument groups so there are some permission things that need to get built
- [x] #108
- [x] #126 
- [x] Styling
- [x] #125 (~~@tarebyte~~)
- [x] Automated Testing of the entire process
- [x] #156 
- [x] Finish all `TODO:`'s in the code
- [x] #164 
- [x] Sockets - to remove a person from the list if they've been challenged while a user is on the page (not high priority. Probably shouldn't mess since release is soon)

atareshawty:
This will probably need to be edited as things continue

atareshawty:
#87 Has to be a part of this

atareshawty:
I'm checking off admin things because of #106

atareshawty:
I checked off `TODOs` because of #157 . It took care of the last one 🎉

atareshawty:
This was a pre-release checklist. We've released. Moving sockets to a separate issue and closing this one

PR #98, opened 2016-07-29
Logging

atareshawty:
YAY LOGGING 🍻 🎉


atareshawty:
Part of #78

atareshawty:
We're using [Paper Trail](https://papertrailapp.com) now! Now all of the logging is just console.logs that are now searchable and last up to 7 days, which will be plenty for what we need

PR #102, opened 2016-08-08
Client js

### LINKED ISSUE #92: More Front End Js
Right now, there's a weird mix between server side redirects and js fetches happening and it's confusing. This takes priority over other things before development can continue so things don't get too hairy

atareshawty:
Fixes #92


db/schema.sql
```diff
@@ -144,24 +144,26 @@ $$ LANGUAGE plpgsql;
 DROP FUNCTION IF EXISTS make_challenge(uId varchar(256), pId int, sId varchar(3));
 CREATE OR REPLACE FUNCTION make_challenge(uId varchar(256), pId int, sId varchar(3))
 RETURNS text as $$
-DECLARE message TEXT; spotOpen boolean; cCount int;
+DECLARE code INT; spotOpen boolean; cCount int;
```
atareshawty: This variable is unnecessary

atareshawty:
- [ ] Checked on staging

PR #116, opened 2016-08-12
Change homepage next performance query

atareshawty:
- [x] Checked on staging

### DIFF
--- a/models/Performance.js
+++ b/models/Performance.js
@@ -122,16 +122,16 @@ module.exports = class Performance {
     });
   }
 
-  findNextWithinWindow() {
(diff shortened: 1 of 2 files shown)


models/Performance.js
```diff
@@ -122,16 +122,16 @@ module.exports = class Performance {
     });
   }
 
-  findNextWithinWindow() {
+  findNextOpenAt() {
```
atareshawty: Not a fan of the function name. Maybe `findNextOrOpenWindow`

atareshawty:
🚀

PR #129, opened 2016-08-20
Current performance

### LINKED ISSUE #128: Check for undefined performance id
There are a few spots where `req.session.currentPerformance.id` is being blindly checked

atareshawty:
Fixes #128 

`Challenges.findAllChallengeablePeopleForUser` checks for performance id and doesn't make a query if not given one.

`Result.findAllForEval` query joins on performances with a closed challenge window


atareshawty:
- [ ] Checked On Staging

PR #142, opened 2016-08-22
Allow the app to run database migrations

tarebyte:
/cc @atareshawty


.eslintignore
```diff
@@ -1,6 +1,7 @@
 experiments/*
 node_modules/*
 bin/*
+db/*
```
atareshawty: Instead of removing the directory from linting, can you fix the errors?

tarebyte: Ugh that's a lot of errors...

atareshawty: Are they auto-generated or copy paste?

atareshawty:
So if I need to make a schema change, do migrations have to be written manually? Also, can you point me to some docs for all of this?

script/setup
```diff
@@ -15,12 +15,25 @@ fi
...
+  echo "Development database already exists, dropping"
+  dropdb challenges_dev
 fi
-script/mock-data
 
+echo "===> Setting up development database..."
+createdb challenges_dev
+
+echo "==> Migrating development"
+./node_modules/db-migrate/bin/db-migrate up --config config/database.json -m db/migrations
```
btmills: npm helpfully symlinks all bins (as long as they're declared as such in the module's `package.json`) to `./node_modules/.bin/bin_name`. Typically this path is preferred to manually referencing the path inside the module, as long as the module declares them.

atareshawty:
@tarebyte  Thank you so much for this! I pulled a few ideas from it into #237 . I'm closing this, but it definitely helped me a lot. I appreciate it 🚀

PR #148, opened 2016-08-28
Change SL eval permission

atareshawty:
# TODO:
- [x] Check on staging


atareshawty:
After my meeting with Chris, he clarified the requirements of what challenges a squad leader should see.
A SL should see the challenges of everyone currently in his/her row and all challenges for a spot in his/row.

For example: Say I'm an R row squad leader ('You're an R row squad leader'). I'd see all challenges involving R13 and R14 along with those pesky E row people challenging for spots into R row.

PR #150, opened 2016-08-28
Password middleware

atareshawty:
# TODO:
- [x] Checked on staging

This was pretty simple to implement

### DIFF
--- a/auth/index.js
+++ b/auth/index.js
@@ -72,7 +72,6 @@ function ensureEvalAbility(req, res, next) {
 }
 
 function ensureNotFirstLogin(req, res, next) {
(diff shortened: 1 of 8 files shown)


auth/index.js
```diff
@@ -70,10 +70,24 @@ function ensureEvalAbility(req, res, next) {
     return res.redirect('/');
   }
 }
+
+function ensureNotFirstLogin(req, res, next) {
+  console.log('Ensure Not First Login', req.user);
```
atareshawty: Remove this debugging log...

atareshawty:
We should reload the page when the password change is successful. For the redirect, it would load the profile page

PR #151, opened 2016-08-28
Force password update

atareshawty:
/cc @tarebyte 

Turns out this was pretty simple


atareshawty:
This was already checked on staging as a part of #150

PR #163, opened 2016-08-31
Time iz hard

atareshawty:
Now storing timestamps as `timestamptz`. I also removed the `current` flag on performances, which was stupid


atareshawty:
This works for the open window, but displays incorrectly on the browser because we're getting local time on the server which is in UTC

atareshawty:
Everything but the display works on staging 🎉

atareshawty:
For the love of God, check this thoroughly on staging

atareshawty:
One more thorough test tomorrow, but I think we're good!

btmills:
Rather than assuming east coast time, can't you ask the browser to display in the user's current time zone?

> I also removed the current flag on performances, which was stupid

Well, "stupid" is one word for redundant :grinning: Watching you critique yourself is a treat.

controllers/Challenges.js
```diff
@@ -7,23 +7,27 @@ const Performance = new models.Performance();
...
         });
-        logger.challengesLog(`${req.user.name} successfully challenged for ${spotId}`);
-        res.json({ code });
-      })
-      .catch((err) => {
-        logger.errorLog('Challenges.create', err);
-        res.status(400).send(err);
-      });
+    } else {
+      return res.json({ code: 4 });
```
btmills: Is `code: 4` defined in an enum somewhere?

middleware/index.js
```diff
@@ -1,41 +1,17 @@
...
-    }
-    next();
-  })
-  .catch((err) => {
-    logger.errorLog('Middleware.currentPerformance', err);
-    next();
-  });
-};
-
 /* eslint-disable consistent-return */
```
btmills: I'm not recently familiar with how Express middleware works. When you `return next()`, does it use the return value from `next` in whatever called the middleware? Why do you sometimes ignore the return value and sometimes `return` it?

controllers/Challenges.js
atareshawty: https://github.com/osumb/challenges/blob/master/db/schema.sql#L126-L145 Is where that gets defined. It's kind of sloppy. `code: 4` will just make an arbitrary error message. It definitely needs to be documented better. The db schema is a bad place for it

middleware/index.js
atareshawty: Good call. I was returning it so eslint would stop yelling at me for inconsistent return. I don't think the return value of the middleware callback affects anything, but a quick read of the docs makes me think that convention is to just call `next()`

db/queries/can-challenge-for-performance.sql
```diff
@@ -7,4 +7,4 @@ SELECT * FROM users WHERE
     NOT EXISTS (SELECT * FROM manage AS m WHERE m.usernamenumber = $1 AND m.performanceid = $2 ORDER BY id DESC LIMIT 1) OR
     (SELECT reason = 'Closed Spot' FROM manage AS m WHERE m.usernamenumber = $1 AND m.performanceid = $2 ORDER BY id DESC LIMIT 1)
   )
-);
+) AND nameNumber = $1;
```
atareshawty: This is out of scope for this PR. Leaving it, but wanted to make it visible

models/Performance.js
```diff
@@ -174,6 +126,6 @@ module.exports = class Performance {
     }
     const now = moment.utc(new Date().toJSON());
 
-    return moment(performance.openat).isBefore(moment(now)) && moment(now).isBefore(moment(performance.closeat));
+    return moment(new Date(performance.openat)).isBefore(moment(now)) && moment(now).isBefore(moment(new Date(performance.closeat)));
```
btmills: This references `openat` and `closeat` properties on `performance`. Both the [schema](https://github.com/osumb/challenges/blob/669aca9df48780e92ecbda663df5598529351b06/db/schema.sql#L261-L262) and the `format` method above return camelCase `openAt` and `closeAt`. Where is this coming from?

btmills:
Trying to give myself a high-level overview of the `Performance` class here. **Comments in bold.**

### Static methods

- `getAttributes()`
- `getIdName()`
- `getTableName`

**Could these three instead be static accessors?** `static get attributes() { return attributes; }`

### Instance methods

- `create(name, performDate, openAt, closeAt)`
    - Adds a new row to the `performances` table.
    - Returns a `Promise` whose value is an object containing `{ id, utcCloseAt }`.
- `findAll(formatString)`
    - Returns a `Promise` whose value is an array of the result of calling the `format` instance method with the row from the database and the `formatString` argument.
- `findNextOrOpenWindow()`
    - Returns the either the next opening performance or the currently open performance. Unlike the `findAll` instance method, it does not call the `format` instance method, instead directly returning the row from the database.
    - **Right now the logic for this is `now() < openAt OR (openAt < now() AND now() < closeAt)`. Note that since you're specifying `ORDER BY openAt`, this is logically equivalent to `now() != openAt AND now() < closeAt`, though I suspect you really want just `now() < closeAt`.**
    - **This query is giving me a weird feeling, so let's dig into it a bit. `closeAt` is a timestamp with time zone, which we know from the client-side implementation should be in UTC. We then compare it to `now()`. So as long as the server's actual timezone matches what it thinks its timezone is, then there should be no issues here. If, say, the server thinks it's in UTC but it's actually running EST (UTC-4), then this comparison could fail.**
- `format(performance, formatString)`
    - Returns an object with some extra data representing a performance.
    - **It's interesting that `closeAt` and `openAt` are strings formatted with the `formatString` argument. I'm curious why dates are being handles as formatted strings on the server and not native `Date` instances, converted to ISO 8601 date strings at the last moment during JSON serialization in the response.**
- `inPerformanceWindow(performance)`
    - Given a performance **(I can't tell whether this is a row from the database or an object from the `format()` instance method.)**, returns whether the current time is between the performance's opening and closing times. **I left a comment inline about the casing of `open(A|a)t` and `close(A|a)t`.**

models/Performance.js
atareshawty: http://stackoverflow.com/questions/21796446/postgres-case-sensitivity

I'm in it too deep now to go back and make everything case sensitive lol

db/queries/can-challenge-for-performance.sql
atareshawty: Actually, remove because of #167

atareshawty:
> Could these three instead be static accessors?

Yes and that feels better

> I suspect you really want just now() < closeAt

Yep! I've been in the weeds too long

> So as long as the server's actual timezone matches what it thinks its timezone is, then there should be no issues here. If, say, the server thinks it's in UTC but it's actually running EST (UTC-4), then this comparison could fail

`now() as time zone 'utc'`?

Overall, the model api isn't very consistent. It's been one of those things that I've mentally just put away because I know how it works. Thanks for pointing it out. This is a great example of the pitfalls of working on a project without anyone else

models/Performance.js
```diff
@@ -151,19 +107,15 @@ module.exports = class Performance {
...
-    const now = moment.utc(new Date().toJSON());
-    const windowOpen =
-      moment(performance.openat).isBefore(moment(now)) &&
-      moment(now).isBefore(moment(performance.closeat));
+    const windowOpen = this.inPerformanceWindow(performance);
 
     return {
-      closeAt: moment.utc(performance.closeat).local().format(formatString),
+      closeAt: moment.utc(performance.closeat).format(formatString),
       current: performance.current,
```
btmills: Didn't you remove `current`?

btmills:
The `Performance` class seems to be the meat of this.

It makes me nervous that there are at least four types of performances running around:

1. The `Performance` class itself, having three static accessor methods, `getAttributes()`, `getIdName()`, and `getTableName()`. (I left a note in the previous comment wondering if these would be better as `static get getAttributes() { /* ... */ }` accessors.)
1. Instances of the `Performance` class, having instance methods `create()`, `findAll()`, `findNextOrOpenWindow()`, `format()`, and `inPerformanceWindow()`.
1. Rows returned from the database.
1. Objects returned from the `format()` instance method.

I like that you've isolated database access in one place. That's a good pattern.

A few things that give me pause here:

1. `format()` and `inPerformanceWindow()` are both instance methods, but they take as an argument a `performance` object of unknown provenance. From context, I can surmise it's a raw database row.
1. You're passing a `formatString` as an argument to `findAll()`, but that doesn't seem related to the query in any way.
1. `format()` converts `openAt` and `closeAt` to be strings. *(I left a note in the comment above about native `Date` instances being more common except on the edges, which would serialize and deserialize ISO 8601 date strings.)*
1. You're using moment.js to handle dates on the backend, even to compare them. That seems like it might obscure what's actually going on. In order to compare `Date` instances, you ought to be able to do `this.openAt < Date.now() && Date.now() < this.closeAt`. Typically one would use moment.js on the frontend for user-friendly parsing, and it does have some handy helper methods (e.g. one week in the future, is leap year). **I'd recommend making the native `Date` instance the canonical timestamp representation within the backend JS** - `inPerformanceWindow()` ought not to need moment.js.

Design pattern wise, a common pattern I've seen would be to have `create()`, `findAll()`, and `findNextOrOpenWindow()` static methods on the `Performance` class. Those would then return `Performance` *instances*. The only instance method would then be `inPerformanceWindow()`, taking no arguments. You may also have some instance accessors like `get  name() { return this._name; }`:

### Static accessors

- `static get attributes()`
- `static get idName()`
- `static get tableName()`

### Static methods

- `static findAll()` - returns a `Promise` whose resolution is an array of `Performance` instances
- `static findNextOrOpenWindow()` - returns a `Promise` whose resolution is `null` or a `Performance` instance

### Instance accessors

- `get name()`
- (Others as necessary)

### Instance methods

- `inPerformanceWindow()` - returns a boolean value indicating whether the sign-up window is currently open
- `toJSON()` - this might be helpful if you need to serialize it to send it to the client

public/js/performances/new.js
```diff
@@ -16,10 +18,10 @@ $('.NewPerformanceSubmit').on('click', () => {
       credentials: 'same-origin',
       method: 'post',
       body: JSON.stringify({
-        closeAt: new Date(inputs.closeAt.val()).toISOString(),
+        closeAt: moment.utc(new Date(inputs.closeAt.val())).format(),
```
btmills: Why the switch to `moment.utc`? The only difference should be the millisecond precision of `toISOString()`, which shouldn't be an issue, should it?

controllers/Users.js
```diff
@@ -88,7 +88,7 @@ function UsersController() {
     if (req.user.admin) {
       Performance.findNextOrOpenWindow()
       .then((performance) => {
-        res.render('users/admin', { user: req.user, performance: Performance.format(performance) });
+        res.render('users/admin', { user: req.user, performance: performance && performance.toJSON() });
```
btmills: You'll have to test this, but I believe `JSON.stringify()` will call `toJSON()` if it is available, so you might be able to just pass `performance` without having to call `toJSON()` explicitly.

models/Performance.js
```diff
@@ -83,97 +70,78 @@ module.exports = class Performance {
...
+      const queryString = 'SELECT * FROM performances WHERE now() < closeAt ORDER BY openAt ASC LIMIT 1';
 
       client.connect();
       client.on('error', (err) => reject(err));
 
       const query = client.query(queryString);
 
-      query.on('row', (result) => {
+      query.on('row', ({ id, name, openat, closeat, performdate }) => {
         client.end();
```
btmills: Depending on how often you're doing database queries, you may find it advantageous to create a helper function to run database queries. It would take a query string and any parameters and handle connecting, running the query, collecting rows, and closing the client when done, and just return a promise. Rough sketch follows.

```js
// Rough sketch
const query = module.exports = (queryString, params) => new Promise((resolve, reject) => {
    const client = utils.db.createClient();
    client.connect(); // Is connect() synchronous? Something seems a bit fishy. Are you creating a new pool every time or just getting a connected client from the pool?
    client.on('error', (err) => reject(err));
    const response = client.query(queryString, params);
    const rows = [];
    response.on('row', (row) => {
        rows.push(row);
    });
    response.on('end', () => {
        client.end();
        resolve(rows);
    });
    response.on('error', (err) => {
       client.end();
       reject(err);
    });
});

// Usage
query('SELECT now();')
.then((rows) => {
    console.log(rows[0].now);
})
.catch((err) => {
    console.error('Whoops!', err);
});
```

atareshawty: I'm going to want to use this in all of the model classes in the same PR. I don't want the scope of this one to expand to more classes. The plan is to make this a part of #166 in some capacity

atareshawty: Along with more refactoring to mimic the changes to the Performance class

controllers/Challenges.js
```diff
@@ -2,42 +2,47 @@ const email = require('../utils').email;
...
-          spotId
-        });
-        logger.challengesLog(`${req.user.name} successfully challenged for ${spotId}`);
-        res.json({ code });
-      })
-      .catch((err) => {
-        logger.errorLog('Challenges.create', err);
-        res.status(400).send(err);
-      });
+    return Performance.findNextOrOpenWindow()
```
btmills: Would it be easier to attach this as `req.currentPerformance` in middleware? I notice you're using this frequently.

atareshawty: There's really no performance bump, correct? I'd rather have it explicitly called by the controllers that need it rather than have it be partially hidden in middleware. If I find that I'm using it more and more (now that the model class is much better 🎉 ) I'll switch to middleware

btmills: Yeah, this would be purely whichever you find easier to use as the developer.

atareshawty:
Just tested this on staging and it all works as expected 🎈 🍻 
@btmills Thank you so much. This is so much better now

PR #167, opened 2016-08-31
Challenge view

atareshawty:
If the user has a result for the challenge, don't display it on the profile


atareshawty:
- [x] Checked on staging

PR #168, opened 2016-09-02
Code cleanup

atareshawty:
# In This PR

- Refactored model classes.
For example:
```js
const { Performance } = require('../models');

Performance.findCurrent() //db fetching methods are now static functions
.then((performance) => {
  /* Do some stuff */
  /* The returned data is an instance of the Performance class */
});
```
All model classes are affected. The other changes are just to reflect the changes in model api

# TODO
- [x] Test pass, but for the love of God, please check this thoroughly on staging!


models/Spot.js
```diff
@@ -3,20 +3,27 @@ const utils = require('../utils');
...
   }
 
-  static getTableName() {
+  static get tableName() {
     return 'spots';
   }
 
   // true is open, false is closed
-  setOpenClose(spotId, open) {
+  static setOpenClose(spotId, open) {
```
btmills: Part of me wants this to me an instance method, so you wouldn't have to pass in the ID. Rough sketch below:

```js
const spot = Spot.findById(123);
spot.open = true;
spot.save();
```

Definitely not a big deal, but I feel like it might encapsulate things a bit more. (Also probably not worth changing before tomorrow.)

models/Result.js
```diff
@@ -31,21 +31,39 @@ const resultsSort = (a, b) => { // eslint-disable-line consistent-return, array-
...
     return 'id';
   }
 
-  static getTableName() {
+  static get tableName() {
     return 'results';
   }
 
-  approve(ids) {
+  static approve(ids) {
```
btmills: You can disregard my comment from earlier. Having these static methods to mutate isn't something I've often seen, but you're consistent, so 👌

models/Result.js
```diff
@@ -333,43 +356,100 @@ module.exports = class Results {
...
+      performanceName: this._performanceName,
+      spotId: this._spotId,
+      winner: this._winnerId === nameNumber
     };
   }
-};
+}
+
+// id, fC, fNN, nA, opponentName, pending, perfDate, perfName, perfId, sC, sNN, spotId, winnerId
+const instanceFromDBRow = ({
```
btmills: I haven't yet figured out a non-annoying way to do this. Let me know if you find one.

atareshawty: This was so annoying to write! I'm glad it's not just me

models/Challenge.js
```diff
@@ -52,6 +60,7 @@ class Challenge {
         resolve(null);
       }
 
+      console.log(user.instrument, user.part, user.spotId, user.nameNumber);
```
btmills: Debug log? Wouldn't be surprised to see this removed in a later commit that I haven't reviewed yet.

auth/index.js
```diff
@@ -27,6 +27,7 @@ passport.deserializeUser((user, done) => {
 });
 
 function isAuthenticated(req) {
+  console.log(req.user);
```
btmills: Debug log

auth/index.js
```diff
@@ -27,7 +27,6 @@ passport.deserializeUser((user, done) => {
 });
 
 function isAuthenticated(req) {
-  console.log(req.user);
```
btmills: Ah yep, I see both of these were later moved (let the record reflect that I anticipated that in my first comment)

models/Result.js
```diff
@@ -430,7 +430,7 @@ class ResultForAdmin {
     this._secondName = sN;
     this._secondNameNumber = sNN;
     this._spotId = spotId;
-    this._winner = winner;
+    this._winnerId = winnerId;
```
btmills: Nice. I like storing the raw data and calculating the result at the place it's needed.

btmills:
Architecturally, this looks solid. I notice you're often using multiple similar classes for different "views" - the <Class>For<View> pattern. I feel like there ought to be a less repetitive way to do that, but given your security requirements, you need to be sure you won't divulge information to the wrong user. I don't have a better idea at the moment, but it's something I'll ponder.

LGTM

PR #170, opened 2016-09-03
Fake data

atareshawty:
Email template and updated fake data for user creation script


atareshawty:
screen shot 2016-09-03 at 7 59 48 pm

atareshawty:
screen shot 2016-09-03 at 8 02 55 pm

Better

PR #172, opened 2016-09-04
User create

atareshawty:


atareshawty:
So this is kind of a one off thing to upload the users and send emails to everyone with their usernames and passwords. **This will not be the official way to upload users into the db for later**

PR #173, opened 2016-09-04
Pg pool

atareshawty:
# In This PR

- Using [pg-pool](https://www.npmjs.com/package/pg-pool)

All models follow the form:
``` js
return Promise((resolve, reject) => {
  db.query(sql, [params...], functionToInstantiateObjectFromDBRow)
  .then(resolve)
  .catch(reject);
});
```

Updates to controllers were made as necessary.

# TODO:
- [ ] Check thoroughly on staging

### DIFF
--- a/models/Challenge.js
+++ b/models/Challenge.js
@@ -27,7 +27,7 @@ class Challenge {
     return new Promise((resolve, reject) => {
       const sql = 'SELECT make_challenge($1, $2, $3)';
 
-      db.query(sql, [userId, spotId, performanceId])
+      db.query(sql, [userId, performanceId, spotId], codeFromRow)
       .then(resolve)
       .catch(reject);
     });
@@ -165,6 +165,8 @@ class ChallengeableUser {
   }
 
 }
+
+const codeFromRow = ({ make_challenge }) => make_challenge;
 
 const instanceFromRowChallenge = ({ name, performanceid, spotid }) =>
   new Challenge(performanceid, name, spotid);
--- a/models/Challenge.js
+++ b/models/Challenge.js
@@ -1,5 +1,5 @@
 const queries = require('../db/queries');
-const utils = require('../utils');
+const { db } = require('../utils');
 
 const attributes = ['id', 'performanceId', 'userNameNumber', 'spotId'];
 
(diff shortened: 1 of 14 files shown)


models/Challenge.js
```diff
@@ -24,156 +24,55 @@ class Challenge {
...
-        resolve(returnCode);
-      });
-
-      query.on('error', (err) => {
-        client.end();
-        reject(err);
-      });
+      db.query(sql, [userId, performanceId, spotId], codeFromRow)
+      .then(resolve)
+      .catch(reject);
```
btmills: Heyyy this is 💯 times more concise

models/Challenge.js
```diff
@@ -24,156 +24,55 @@ class Challenge {
...
-      });
-
-      query.on('end', () => {
-        client.end();
-        resolve(users);
-      });
-
-      query.on('error', (err) => reject(err));
+      db.query(sql, [user.instrument, user.part, user.spotId, user.nameNumber, performanceId], instanceFromRowChallengeableUser)
+      .then(resolve)
```
btmills: Instead of wrapping this in return new promise, just return the promise directly, no then resolve catch reject. Above instead of resolve null, return Promisw.resolve(null)

atareshawty: I didn't think it could get simpler

btmills: Saved you a few more lines ❌

models/Challenge.js
```diff
@@ -26,155 +26,38 @@ class Challenge {
...
-        client.end();
-        resolve(returnCode);
-      });
-
-      query.on('error', (err) => {
-        client.end();
-        reject(err);
-      });
-    });
+    return db.query(sql, [userId, performanceId, spotId], codeFromRow);
```
btmills: This is amazing. ‼️

models/Challenge.js
```diff
@@ -26,155 +26,38 @@ class Challenge {
...
-      query.on('error', err => {
-        client.end();
-        reject(err);
-      });
-    });
+    return db.query(sql, [performanceId], instanceFromRowChallengeForCSV);
   }
 
   static findAllRawForPerformance(performanceId) {
     const sql = 'SELECT * FROM challenges WHERE performanceId = $1';
```
btmills: Why do you include the SQL in some methods and extract it in others?

atareshawty: It just depended on how complicated the SQL was. Something simple like is fine as a string, but I preferred to write the SQL for bigger queries

models/Manage.js
```diff
@@ -26,56 +26,23 @@ module.exports = class Manage {
...
   static create(attributes) {
-    return new Promise((resolve, reject) => {
-      const client = db.createClient();
-      const { sql, values } = db.queryBuilder(Manage, attributes);
-
-      client.connect();
-      client.on('error', (err) => reject(err));
-
-      const query = client.query(sql, values);
+    const { sql, values } = db.queryBuilder(Manage, attributes);
```
btmills: Note to self: Based on the name alone, I want to check the `buildQuery` helper for SQL injection vulnerabilities.

atareshawty: See `spec/query-builder-spec.js`

models/Manage.js
```diff
@@ -26,56 +26,23 @@ module.exports = class Manage {
...
-        resolve(actions);
-      });
-
-      query.on('error', (err) => {
-        client.end();
-        reject(err);
-      });
+      db.query(sql, [performanceId], instanceFromRowManage)
+      .then(resolve)
+      .catch(reject);
```
btmills: You can unwrap this too

atareshawty: Darn! I missed one

models/Performance.js
```diff
@@ -27,95 +27,37 @@ module.exports = class Performance {
...
-          client.end();
-          reject(err);
-        });
-      }
-    });
+    const sql = 'INSERT INTO performances (name, performDate, openAt, closeAt) VALUES ($1, $2, $3, $4) RETURNING id, closeAt';
+
+    if (
+      Number.isNaN(Date.parse(openAt)) ||
+      Number.isNaN(Date.parse(closeAt)) ||
```
btmills: This is an interesting pattern for date validation that I'd not seen before. I'll have to look into how it works later

models/Performance.js
```diff
@@ -27,95 +27,37 @@ module.exports = class Performance {
...
+      Number.isNaN(Date.parse(performDate))
+    ) {
+      return Promise.reject();
```
btmills: It's helpful to return a reason for the rejection (I know, this is just old code that got moved in the diff, you can handle this later)

models/Result.js
```diff
@@ -280,78 +133,38 @@ class Result {
...
-      });
+        db.query(switchSql, [filteredResultsIds])
+        .then(() => {
+          db.query(oneUserSql, [onePersonResultIds])
+          .then(resolve)
+          .catch(reject);
+        })
+        .catch(reject);
+      })
+      .catch(reject);
```
btmills: More unwrapping available here 🎁

atareshawty: I'm not in a good enough state to unwrap those nested promises and that's a pretty important part of the code base. I'm going to save that one for later 😜

models/Result.js
```diff
@@ -280,78 +133,38 @@ class Result {
...
-        reject(new Error('No id provided with attributes'));
-      }
+    if (typeof id === 'undefined') {
+      return Promise.reject(new Error('No id provided with attributes'));
+    }
 
-      delete attributes.id;
-      const { sql, values } = db.queryBuilder(Result, attributes, { statement: 'UPDATE', id });
-      const query = client.query(sql, values);
+    delete attributes.id;
```
btmills: You're mutating a passed argument here. I don't know where it comes from, could this get you in trouble elsewhere? Might be safer to copy needed props out to a new object. Even if it's just a lodash.omit(['id']); or even const copied = Object.assign({}, attributes); delete copied.id;

models/Result.js
```diff
@@ -280,78 +133,38 @@ class Result {
...
-        resolve();
-      });
-
-      query.on('error', (err) => {
-        client.end();
-        reject(err);
-      });
+      db.query(sql, [comments1, comments2 || null, winnerId, firstNameNumber])
+      .then(resolve)
+      .catch(reject);
```
btmills: It's like it's your birthday over here 🎁

models/Result.js
btmills: Haha that's probably wise

models/User.js
```diff
@@ -35,117 +35,40 @@ class User {
...
-      });
-
-      query.on('error', (err) => {
-        reject(err);
-        client.end();
-      });
-    });
+  static canChallengeForPerformance(user, performanceId) {
+    if (!performanceId) {
+      return Promise.resolve(false);
```
btmills: 1) Will performanceId ever be 0?
2) If performanceId is not passed, is that an indication of a programming error, or is it something that could happen normally? If the former, that would make this a rejection.

1 may be a ship stopper, 2 almost certainly isn't

models/User.js
```diff
@@ -35,117 +35,40 @@ class User {
...
-        client.end();
-        resolve(password);
-      });
-      query.on('error', (err) => {
-        client.end();
-        reject(err);
-      });
-    });
+    const sql = 'UPDATE users SET password = $1, new = false WHERE nameNumber = $2';
+    const password = bcrypt.hashSync(newPassword, bcrypt.genSaltSync(1)); // eslint-disable-line no-sync
```
btmills: Later, not now, make this async. Leaving a note just to call it out

models/User.js
atareshawty: Performance is an auto-incrementing db table, so it should never be 0

models/User.js
```diff
@@ -160,87 +83,22 @@ class User {
...
-        resolve(null);
-      });
-
-      query.on('error', (err) => {
-        client.end();
-        reject(err);
-      });
+      db.query(sql, [id], instanceFromRowUser)
+      .then(([user]) => resolve(user))
+      .catch(reject);
```
btmills: Unwrap this with `return db.query(...).then(([user]) => user);`

utils/return-argument.js
```diff
@@ -0,0 +1 @@
+module.exports = (arg) => arg;
```
btmills: This is called the identity function

package.json
```diff
@@ -41,6 +41,7 @@
     "passport": "^0.3.2",
     "passport-local": "^1.0.0",
     "pg": "^4.5.1",
+    "pg-pool": "^1.4.3",
```
btmills: If you have pg-pool, do you still need pg?

btmills: Ah, still using pg for createClient below. I assume createClient is used places and not just the query helper?

utils/db.js
```diff
@@ -1,9 +1,32 @@
...
+
+const createClient = () => new pg.Client(config.db.postgres);
 
-const createClient = () =>
-  new pg.Client(config.db.postgres);
+const query = (sql, params, parseRow = returnArgument) =>
+  pool.query(sql, params)
+  .then((data) => (data.rows || []).map((row) => parseRow(row)))
+  .catch((err) => {
+    throw err;
```
btmills: This catch handler is redundant - throwing in a promise will cause a rejection, so this is catching just to pass through the rejection from before - you can take this out and it will still propagate

package.json
atareshawty: Create client is used for some one off db connections. Not for anything user facing

atareshawty:
Checks out on staging!

atareshawty:
Thanks @btmills

PR #175, opened 2016-09-04
Script to just set up the db schema

atareshawty:
Again, self explanatory


script/schema
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+# script/mock-data: Populate db with fake-data
```
atareshawty: Wrong comment...

PR #191, opened 2016-09-09
Manual spot change

atareshawty:
# In This PR

- Manually updating user spot id incase things get weird (things have already gotten weird)

### DIFF
--- a/models/User.js
+++ b/models/User.js
@@ -85,8 +85,12 @@ class User {
     return db.query(sql, [id], instanceFromRowUser).then(([user]) => user);
   }
 
-  static findAll() {
-    const sql = 'SELECT * FROM users';
+  static indexMembers() {
+    const sql = `
+      SELECT * FROM users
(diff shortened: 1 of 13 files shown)


models/User.js
```diff
@@ -85,8 +85,12 @@ class User {
     return db.query(sql, [id], instanceFromRowUser).then(([user]) => user);
   }
 
-  static findAll() {
-    const sql = 'SELECT * FROM users';
+  static indexMembers() {
+    const sql = `
+      SELECT * FROM users
+      WHERE NOT spotId IS NULL
```
atareshawty: Should be checking on role. If a user doesn't have a spot, then they won't show up

atareshawty:
- [ ] Checked on staging

atareshawty:
Should definitely add logging for these actions

models/User.js
atareshawty: Just kidding. If they don't have a spotId, the order by will error out

PR #195, opened 2016-09-11
Index results

atareshawty:
- [x] Checked on staging


views/results/index.handlebars
```diff
@@ -0,0 +1,27 @@
+<div class="ResultsIndex">
+{{#each performanceResultsMap}}
+  <h1>{{ this.performanceName }}</h1>
+    <div class="ResultsApproveList">
```
atareshawty: 👍  for reusing CSS classes

atareshawty: Might want to rename it in the future though

PR #198, opened 2016-09-12
Add migrations!

kylekthompson:
Here's what happened:
1. Switched to a `challenges_test` database for testing
2. Update seed files to use `db.query` instead of a `pg` client
3. Update seed files to use Promises
4. Move `schema.sql` into `up` and `down` migrations in `db/migrations`
5. Add `ON DELETE CASCADE` to references on the tables (so we can `DELETE FROM` in `reset.sql`)
6. Update `script/mock-data` to be `script/seed` which runs a reset and seed
7. Add migrations (using Knex and a script to create the templates)

The NPM scripts:
- `seed`: replacement for `mock-data`. clears the database and then fills it with the original seed data
- `migrate:current`: shows the current migration state of the database (useful to find out what hasn't run)
- `migrate:latest`: runs migrations all migrations (puts the database in the most recent state)
- `migrate:make`:  usage: `npm run migrate:make addSomeNewFeature` -- creates the migration template files
- `migrate:rollback`: rolls back the most recent batch of migrations. Note: this is not the most recent _migration_ but the batch of migrations (i.e. if you make 3 migrations, then run `migrate:latest`, then run `migrate:rollback` it will roll 3 migrations back)

This will require quite a bit of planning before merging into master. Since this migration needs to be the base, the database would need dropped, and that would remove all historical data (no good)!

Another option would be to keep the gigantic `schema.sql` file and just make a migration with the updates. Unfortunately then we would still have the file and another step in `script/setup`.

Let me know if you need anything else (documentation or updates)!

/cc @atareshawty


kylekthompson:
@atareshawty just a heads up: migrations won't run on Heroku with this current setup. I just ran into the issue when setting up Heroku for my stuff.

The `DATABASE_URL` supplied by Heroku needs to be parsed first. It looks like you are already doing this in `utils/db.js`. I'll do a little refactoring to make that parsing reusable later today and fix the migrations on Heroku.

atareshawty:
Thanks for doing this! I definitely want to get this merged, but it might be a while. This is a pretty big change while the prod database is being used. It's currently in a stable state, so I don't want to rock the boat.

atareshawty:
@kylekthompson Thank you so much for this! I pulled a few ideas from it into #237 . I'm closing this, but it definitely helped me a lot. I appreciate it 🚀

PR #202, opened 2016-09-12
Manual part change

atareshawty:
## ~~This is a working PR~~

# TODO:
- [x] Actually implement UI for manual part changes
- [x] Check on staging

### DIFF
--- a/controllers/Users.js
+++ b/controllers/Users.js
@@ -158,7 +158,7 @@ function UsersController() {
     User.update(nameNumber, req.body)
     .then(() => res.json({ success: true }))
     .catch((err) => {
-      console.error(err);
+      logger.errorLog('Users.update', err);
       res.json({ success: false });
     });
   };
--- a/controllers/Users.js
+++ b/controllers/Users.js
@@ -14,8 +14,8 @@ function UsersController() {
     const { oldPassword, newPassword } = req.body;
 
     if (bcrypt.compareSync(oldPassword, req.user.password)) { // eslint-disable-line no-sync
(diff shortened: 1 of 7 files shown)


controllers/Users.js
```diff
@@ -163,6 +149,19 @@ function UsersController() {
...
+
+  this.update = (req, res) => {
+    const { nameNumber } = req.body;
+
+    delete req.body.nameNumber;
+
+    User.update(nameNumber, req.body)
+    .then(() => res.json({ success: true }))
+    .catch((err) => {
+      console.error(err);
```
atareshawty: Need to use `logger`

public/js/users/index-member.js
```diff
@@ -1,17 +1,85 @@
...
+$(editSpotButtonClass).on('click', (e) => {
+  editSpotClickEventFunction(e);
 });
 
-const clickEventFunction = (e) => {
+$(editPartButtonClass).on('click', (e) => {
+  editPartClickEventFunction(e);
+});
+
+const editPartClickEventFunction = (e) => {
```
atareshawty: If another button needs added to change stuff, strongly consider refactoring this. There's some reusable code here

atareshawty:
It's annoying that the existing part isn't the first option in the select

atareshawty:
> It's annoying that the existing part isn't the first option in the select

It *was* annoying

atareshawty:
Looks good on staging. I'm going to sleep on reviewing this code though

atareshawty:
Also fixed bug in the change password backend flow

PR #203, opened 2016-09-13
Destructure array in results controller

atareshawty:
Fixes bug where results get switched every time results get approved


atareshawty:
Although the results were switched correctly, it was premature

PR #205, opened 2016-09-13
Results approve

atareshawty:
Actual fix for what #203 was trying to do


atareshawty:
Switching spots should have action logging for before and after

atareshawty:
> Switching spots should have action logging for before and after

Done!

atareshawty:
- [ ] Actually checked on staging

PR #208, opened 2016-09-13
Change performanceResultMap into an array rather than plain object

### LINKED ISSUE #204: Change Sorting on Previous Challenge Results
Really quick change

atareshawty:
Fixes #204


atareshawty:
- [x] Checked on prod, actually. That was bad, but that's the only place I could make sure it was correct with the data I had

PR #211, opened 2016-09-22
Give min height for .ResultsEvaluation

atareshawty:
This fixed some issue on prod that has somehow presented itself after it worked the first two times


atareshawty:
Checked on staging on my iPad for mobile safari

PR #213, opened 2016-09-28
Scheduled job

### LINKED ISSUE #199: Better Job Scheduling
Node Scheduler keeps the job in memory. So when the server restarts, we lose the scheduled jobs. Heroku restarts it's dynos every 24 hours. The server(s) restart on deploys. Canceling jobs happens right now by just manually restarting the dynos.

I need a solution that:
- [ ] Can programmatically schedule jobs to send the challenge list email
- [ ] Can stay scheduled through things like node restarting/deploys
- [ ] Can manually be altered through UI/terminal (preferably UI)

atareshawty:
This is a little more than job scheduling

- Added new query for the creation of empty results. This way, we can run the job again without duplicating results
- Reorganized the jobs directory a little. Only tests and `clock.js` are touching it, so the directory structure it a little different from the rest of the modules. This is to discourage using `jobs` as a module outside of the clock.js unless you **really** need to.
- Performances controller no longer schedules the email job 🎉  I'm so glad this is gone!
- There were some necessary additions to the performance and result model to make this work (no existing functionality was change)

This is running on staging now with success 

Fixes #199


atareshawty:
**NOTE: The clock process will need to be scheduled by one dyno. `heroku ps:scale clock=1 --remote prod`**

atareshawty:
I checked this on staging by doing the following:

- Running `heroku ps:scale clock=1 --remote staging`
- Deploying the current branch (started the clock process from the Procfile)
- Creating a performance

Because the time of the window closing was a multiple of 5, I got the email right when the performance closed. It has been checked a few more times after and I haven't gotten emails, so it seems like we're good to go!

jobs/clock.js
```diff
@@ -0,0 +1,10 @@
+const cron = require('node-cron');
+const emailChallengeList = require('./email-challenge-list.js');
+
+// Every 5 minutes, from 8:00AM - 7:00 PM ETC from Monday through Friday
```
btmills: Will signups ever close on a Sunday?

atareshawty: I don't remember that ever happening

atareshawty: I've been going back and fourth on how long the job should run. For full flexibility, it probably should just be between 8:00 AM - 7:00PM everyday

atareshawty: Changed. It runs from 12 - 23 UTC everyday. Which is the equivalent of 8:00AM - 7:00PM EST

PR #216, opened 2016-10-03
Send challenge list button

atareshawty:
This fixes a lot of bad that accompanied the broken scheduling. This will be the general framework for how performance editing works. Right now, it's just a route (with no link in the UI) to email the challenge list.

For right now, only I'll know this is an option. When editing is possible, I'll add the link in the navbar


atareshawty:
This works on staging

atareshawty:
> This works on staging

And safari. This is :shipit:

PR #218, opened 2016-10-10
Use a single media query instead of two

kylekthompson:
In `Header`, you can use one media query instead of two.


atareshawty:
Ok. I have finally have the react app in a state where it's good to merge this!

PR #219, opened 2016-10-13
React

atareshawty:
# In This PR

- We have React now! 🍻 🎉 

Because of this, a few things were changed server side:

- Revamped routing (this needed done anyway)
- Moved from session auth to jwt's (also needed done anyway)

# TODO:
- [x] production webpack config
- [x] Test thoroughly on staging!
- [x] Don't render `/login` if logged in. Redirect home

Since this isn't going into master, the entire app's worth of React isn't implemented yet and a lot of code needs to die still.

In retrospect, I probably should have broken up this PR even more...


atareshawty:
This actually worked really freaking well on staging! *knocks on wood* (on the first try, even!)
The auth problem with the 'iOS webapp' is fixed with JWT

auth/index.js
```diff
@@ -1,106 +1,52 @@
-/* eslint-disable callback-return */
-const passport = require('passport');
-const Strategy = require('passport-local').Strategy;
-const models = require('../models');
-const User = models.User;
-const bcrypt = require('bcrypt');
+/* eslint-disable callback-return, no-unused-expressions */
```
atareshawty: Might be able to get rid of this line

config/index.js
```diff
@@ -2,8 +2,8 @@ module.exports = {
   db: {
     postgres: process.env.DATABASE_URL || 'postgres://localhost:5432/challenges_dev'
   },
-  passport: {
-    secret: process.env.PASSPORT_SECRET || 'keyboard cat'
+  auth: {
+    secret: process.env.PASSPORT_SECRET || 'notMuchOfASecret'
```
atareshawty: Should change this name since we're not using Passport anymore more

controllers/Auth.js
```diff
@@ -0,0 +1,36 @@
...
+        const userJSON = user.toJSON();
+
+        delete userJSON.password;
+
+        const token = jwt.sign(userJSON, auth.secret);
+
+        res.json({ token });
+      })
+      .catch((err) => {
+        console.log('as;dlkjasdf;lkj');
```
atareshawty: ...lol kill this with 🔥

package.json
```diff
@@ -8,6 +8,7 @@
   },
```
atareshawty: Not in this PR, but I should really go through here and make sure everything is kosher. For example, there's probably some packages that aren't being used and some deps that would be in dev-deps

public/components/navbar.jsx
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable react/no-multi-comp */
```
atareshawty: This line can be removed

public/components/not-found.jsx
```diff
@@ -0,0 +1,22 @@
...
+import { Link } from 'react-router';
+import Media from 'react-media';
+
+import './not-found.scss';
+
+const minWidth = 768;
+
+const NotFound = () => (
+  <div className="NotFound">
+    <Media query={{ minWidth }}>
```
atareshawty: This media query is no longer needed

public/components/not-found.scss
```diff
@@ -0,0 +1,9 @@
+.NotFound {
+  align-items: center;
+  display: flex;
+  flex-direction: column;
+}
+
+.NotFound-image--mobile {
```
atareshawty: This won't be needed anymore

public/index.jsx
```diff
@@ -0,0 +1,65 @@
...
+};
+
+const App = () => {
+  return (
+    <BrowserRouter>
+      {
+        ({ router }) => (
+          <div className="App">
+            <Header />
+            <Navbar onLogout={handleLogout.bind(null, router)} />
```
atareshawty: Components that need access to the user, should be given the user by this component. Examples right now would be `Navbar` and `Profile`

public/utils/auth.js
```diff
@@ -0,0 +1,30 @@
+import 'whatwg-fetch';
```
atareshawty: Remove this

auth/index.js
atareshawty: We can't...

controllers/Auth.js
atareshawty: But add error logger to the catch

PR #221, opened 2016-10-18
Profile page

atareshawty:
Recreating the profile page!

I added a few styling modifications to make mobile better


atareshawty:
- [ ] Checked on Staging

PR #222, opened 2016-10-18
Challenge select

atareshawty:
Challenge Select In React!


public/components/challengeable-users.jsx
```diff
@@ -0,0 +1,62 @@
+import React, { Component, PropTypes } from 'react';
+
+import './challengeable-users.scss';
+
+const getSelectOptionFromChallengeableUser = (challengeableUser) => {
```
atareshawty: This should really be another component. It's essentially one now

atareshawty:
- [ ] Tested on staging

PR #224, opened 2016-10-20
Results evaluate

atareshawty:
- [x] Checked on staging


atareshawty:
This probably needs a banner for successful evals

atareshawty:
> This probably needs a banner for successful evals

Done!

PR #226, opened 2016-10-23
Past results

atareshawty:
⚠️  Test this again on staging ⚠️ 

It seemed dumb to write multiple Result like components, so I made a generic one that takes an optional button prop. I also renamed the result component being used for profile to avoid naming issues

- [x] Checked on staging


atareshawty:
Need to have text for no previous results

atareshawty:
> Need to have text for no previous results

Done

atareshawty:
This works exactly the same as it did on staging!

PR #227, opened 2016-10-24
Create performance

atareshawty:
- [x] Checked on staging

### DIFF
--- a/public/components/create-performance.jsx
+++ b/public/components/create-performance.jsx
@@ -1,0 +1,110 @@
+import React, { Component } from 'react';
+import Datetime from 'react-datetime';
+import moment from 'moment';
+
+import '../../node_modules/react-datetime/css/react-datetime.css';
+import { api } from '../utils';
+import Banner from './banner';
+
+const inputValid = (name, date, openAt, closeAt) => {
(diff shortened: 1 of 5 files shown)


atareshawty:
The `react-datetime` picker is 🔥

public/components/create-performance.jsx
```diff
@@ -0,0 +1,111 @@
...
+  if (!moment(date).isValid() || !moment(openAt).isValid() || !moment(closeAt).isValid()) return false;
+  return true;
+};
+
+export default class CreatePerformance extends Component {
+
+  constructor() {
+    super();
+    this.state = {
+      closeAt: '',
```
atareshawty: State is getting set here and three variables aren't actually doing anything

public/components/create-performance.jsx
```diff
@@ -0,0 +1,111 @@
...
+      date: '',
+      openAt: '',
+      success: false,
+      valid: true
+    };
+    this.handleCloseAtChange = this.handleCloseAtChange.bind(this);
+    this.handleClick = this.handleClick.bind(this);
```
atareshawty: If we're trying to alphabetize, this one is out of order

PR #229, opened 2016-10-25
Auth bug

atareshawty:
Whoops...

### DIFF
--- a/controllers/Auth.js
+++ b/controllers/Auth.js
@@ -13,9 +13,11 @@ class Auth {
       .then((user) => {
         if (!user) {
           res.status(404).send();
+          return;
         }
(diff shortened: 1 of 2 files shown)


controllers/Auth.js
```diff
@@ -13,9 +13,11 @@ class Auth {
       .then((user) => {
         if (!user) {
           res.status(404).send();
+          return;
         }
-        if (!bcrypt.compareSync(password, user.password)) { // eslint-disable-line no-sync
+        if (!bcrypt.compareSync(password, (user && user.password) || '')) { // eslint-disable-line no-sync
```
atareshawty: Actually, now that we're returning, this line won't be needed

PR #232, opened 2016-11-03
Forgot password

atareshawty:
# Flow for when a user inevitably forgets his/her password.

## FLOW:

- User clicks `Forgot Password?` link on login screen
- At Forgotten Password page, user enters name.number and email for recovery
- If name.number exists and matches email, an email will be sent with link for reset
- Link has a unique id
- On reset page, if the `password_change_request` object is found for that id, and it hasn't expired or used, the user can enter a new password (twice) to reset
- If successful, a link will send them home so they can log in


atareshawty:
This works on staging 🎉

atareshawty:
The scope of this PR is broadening a little bit. This is going to include a 're-do' of auth

atareshawty:
This has broadened too much for this current PR, so I'm closing

PR #233, opened 2016-11-10
React cleanup

atareshawty:
Alright, there's a bunch of stuff going on here... Sorry, self.

First off, auth has changed a little. Tokens can now expire, and revoked (but no on an individual level). On every successful request, a token gets sent with an updated expires date. Rather than send a new token in every controller function, I added middleware in `app.js` that sends the token along with the json response.

The flow is as follows:

- Request is received by server
- Token is checked for expiration/revocation
    - If successful, new token is attached to `res.locals.token`, `next()` is called
- Router gets request and chooses the correct controller action
- Controller gets request and does whatever it needs to do. Attaches it's response to `res.locals.jsonResp` and calls `next()`
- Final middleware at `/api` actually calls `res.json()` with `res.locals.token` and `res.locals.jsonResp`

I added password resetting.

That flow is as follows:

- User heads to `/forgotPassword`
- Enters name.# and email to send password reset
- Request is made and if successful, the user will be sent a unique link with some hashed id associated with the request
- They'll enter a new password (twice) and be sent back to login when successful.
- **After a password is reset, all tokens associated with the user will no longer be valid. That's handled on the `revoke_token_date` user field**

For some reason, an `ApiWrapper` component was added in this PR. It takes care of the `componentDidMount` logic in components that fetch data on mount

For some reason, I wanted to add more pain to this PR, so I revised the `canChallengeForPerformance` query so it's more accurate


controllers/Auth.js
```diff
@@ -16,21 +15,14 @@ class Auth {
           return;
         }
         if (!bcrypt.compareSync(password, user.password)) { // eslint-disable-line no-sync
-          res.status(404);
+          res.status(404).send();
           return;
```
atareshawty: Probably don't need this return

controllers/Users.js
```diff
@@ -10,30 +10,36 @@ const Spot = models.Spot;
...
+        performance: performance && performance.toJSON(),
         results: results.map((result) => result.toJSON()),
-        performance: performance && performance.toJSON()
-      });
+        spotId: dbUser.spotId
+      };
+      next();
     })
     .catch((err) => {
+      console.trace(err);
```
atareshawty: Get rid of this

public/components/forgot-password.jsx
```diff
@@ -0,0 +1,98 @@
...
+
+    return (
+      <div className="ForgotPassword">
+        {badInput && <Banner message="Please enter both your dot number and email" />}
+        <h3>Enter your name.number and email to recover your email</h3>
+        <div className="ForgotPassword-inputs">
+          <div className="ForgotPassword-input">
+            <label className="ForgotPassword-inputs-label">name.#</label>
+            <input
+              autoCorrect="off"
```
atareshawty: `autoFocus`

public/components/reset-password.jsx
```diff
@@ -0,0 +1,149 @@
...
+
+    if (loading) {
+      return <div>Loading...</div>;
+    }
+
+    if (success) {
+      return (
+        <div>
+          <h2>Your password has been reset!</h2>
+          <h3><Link to="/login">Go Home</Link></h3>
```
atareshawty: Change this to `/login`

public/utils/auth.js
```diff
@@ -28,16 +28,32 @@ const authenticate = (nameNumber, password) =>
...
+
 const getUser = () => localStorage[LOCAL_STORE_STRING] && jwtDecode(localStorage[LOCAL_STORE_STRING]);
 
-const isAuthenticated = () => typeof getUser() !== 'undefined';
+const isAuthenticated = () => {
+  const user = getUser();
+
+  if (!user) {
+    return false;
+  } else if (user.revokeTokenDate < user.iat || user.expires < new Date().getTime()) {
```
atareshawty: The first condition check is backwards

atareshawty:
Check this thoroughly on staging!

controllers/Auth.js
atareshawty: YES WE DO!

atareshawty: ............................

atareshawty:
This all checks out on staging ✅

PR #237, opened 2016-12-15
Seed Data

atareshawty:
Much less going on compared to the last one!

I added seed data 🎉 

There used to be a disgusting process to get the mock data from the spread sheets. I was thinking about getting rid of the sheets all together, but I can't really have randomly generated data for testing. The seed script reads directly from the data and loads it. The code is a lot cleaner.

I also added migrations 🍻 

I converted the most recent `schema.sql` into the initial migration up file. Instructions on how to run them are updated in docs


db/database.json
```diff
@@ -0,0 +1,14 @@
+{
+  "development": {
+    "driver": "pg",
+    "host": "localhost",
+    "user": "",
+    "password": "",
+    "database": "challenges_dev"
+  },
+  "test": {
```
atareshawty: Just remove this until the testing PR

spec/fixtures/utils/index.js
```diff
@@ -1,132 +1,75 @@
...
-//   getTestUsersFromExcelFile,
-//   getSpotsFromExcelFile,
-//   getMockChallengesList,
-//   getFakeResults
-// }
+const path = require('path');
+const xlsx = require('node-xlsx');
+const bcrypt = require('bcrypt');
+
+const parseChallengeListFromExcelFile = (type) => {
```
atareshawty: THIS IS SO MUCH BETTER

PR #238, opened 2016-12-15
Fix all of the tests!

atareshawty:
If this works on the first try, I will shit myself.

Basically, we're loading creating the test database, loading test data (similar to #237 ), running tests, and dropping the test database. We're no longer effecting the seed data when we run tests 🎉 🍻  Thank god


atareshawty:
CI actually passes, so LGTM 🚀

PR #242, opened 2016-12-18
Challenge select

atareshawty:
# Challenge Select

- Added migration to add `delete_challenge` function
- Added `ChallengeSelect` component

Result: A user can now undo their challenge after making a selection 🎉


atareshawty:
I also fixed the deploy script and checked this on staging. It works

PR #243, opened 2016-12-20
Evaluate results

atareshawty:
Made the back end better and the front end looks like this:

screen shot 2016-12-19 at 8 47 06 pm

screen shot 2016-12-19 at 8 47 19 pm


client/src/challenge/challenge-evaluations.jsx
```diff
@@ -0,0 +1,124 @@
...
+import ApiWrapper from '../shared-components/api-wrapper';
+import UserComments from './user-comments';
+import UserComparison from './user-comparison';
+
+const endPoint = '/results/evaluate';
+const postEndPoint = '/results/evaluate';
+
+class ChallengeEvaluations extends Component {
+
+  static propTypes() {
```
atareshawty: Make this an instance property of the class

client/src/challenge/challenge-evaluations.jsx
```diff
@@ -0,0 +1,124 @@
...
+  handleEvaluationToggle() {
+    const { evaluating, results } = this.state;
+
+    if (evaluating) {
+      this.setState({
+        evaluating: !evaluating,
+        results: results.map(({ id, ...rest }) => ({
+          ...rest,
+          id,
+          firstComments: this.refs[id].value().firstComments,
```
atareshawty: I want this to be an instance variable of the class. So the call would look like `this.refs[id].value.firstComments`. Then in the class
```js
class UserComments extends Component {
...
get value() {
   return this.state;
}
...
}
```

client/src/challenge/challenge-evaluations.jsx
```diff
@@ -0,0 +1,124 @@
...
+
+    if ((results || []).length < 1) {
+      return (
+        <div className="ChallengeEvaluations">
+          <h2>There are no challenges for you to evaluate</h2>
+        </div>
+      );
+    }
+
+    if (evaluating) {
```
atareshawty: This can be condensed a lot. Do that

client/src/challenge/user-comparison.jsx
```diff
@@ -0,0 +1,122 @@
...
+import { Card, CardActions, CardTitle, CardText } from 'material-ui/Card';
+import classnames from 'classnames';
+import Dialog from 'material-ui/Dialog';
+import FlatButton from 'material-ui/FlatButton';
+
+import './user-comparison.scss';
+
+export default class UserComparison extends Component {
+
+  static propTypes() {
```
atareshawty: Make this an instance property of the class

atareshawty: `static get propTypes() { ... }`

PR #258, opened 2017-02-02
Fix react router

atareshawty:
React router change since their alpha release. 'Bleeding edge', yay.

I also used this as an opportunity to organize the `Fetch` component. The whole `apiWrapper` thing was getting out of hand, so I made it cleaner


atareshawty:
It works on both prod and staging

PR #259, opened 2017-02-02
Edit performance

atareshawty:
List all performances for editing


api/controllers/performances-controller.js
```diff
@@ -18,17 +18,30 @@ class PerformancesController {
     });
   }
 
-  static index(req, res) {
+  static index(req, res, next) {
     Performance.findAll('MMMM Do, h:mm:ss a')
```
atareshawty: `Performance.findAll` no longer takes any arguments

api/controllers/performances-controller.js
```diff
@@ -18,17 +18,30 @@ class PerformancesController {
...
+      next();
     })
     .catch((err) => {
       logger.errorLog('Performances.index', err);
       res.render('static-pages/error', { user: req.user });
     });
   }
 
+  static update({ body }, res, next) {
+    Performance.update(body)
```
atareshawty: This needs data validation, but that's something that's in the pipeline. Not in the scope of this PR

client/src/performance/performance.jsx
```diff
@@ -0,0 +1,150 @@
...
+              <Datetime
+                onChange={this.handlePerformDateChange}
+                textFormat={performDateFormat}
+                value={performDate}
+              />
+            </div> :
+            `Performance on ${moment(performDate).format(performDateFormat)}`
+          }
+        />
+        {editing ?
```
atareshawty: We should be checking for `listExported` to change the open and close times. If the list is exported, a performance shouldn't be exported

client/src/performance/performance.jsx
```diff
@@ -0,0 +1,150 @@
...
+              />
+            </div>
+          </CardText> :
+          <CardText>
+            <div>Opens at: {moment(openAt).format(openCloseDateFormat)}</div>
+            <div>Closes at: {moment(closeAt).format(openCloseDateFormat)}</div>
+          </CardText>
+        }
+        <CardActions>
+          {!listExported && (editing ?
```
atareshawty: Remove this `listExported` check. Name and performance date can still be edited, regardless of whether or not it was exported already

atareshawty:
[ ] Tested on staging and prod

atareshawty:
This is so gross

screen shot 2017-02-02 at 5 04 44 pm

atareshawty:
screen shot 2017-02-14 at 2 47 48 pm

Better 🎉

atareshawty:
This was tested on staging/prod, but this code has been sitting for a while. I need to pass through this again

atareshawty:
Moving to Rails, this is no longer relevant

PR #262, opened 2017-02-20
Rails, because why not?

atareshawty:
This (unfortunately) giant PR is the first step in migrating the api server to rails. This is a good thing, I swear!


test/models/spot_test.rb
```diff
@@ -0,0 +1,340 @@
...
+    spot = build(:spot, row: nil)
+    refute spot.valid?
+  end
+
+  test 'spot must have file' do
+    spot = build(:spot, file: nil)
+    refute spot.valid?
+  end
+
+  test 'a and x rows should allow efer, solo, first trumpet parts' do
```
atareshawty: this test can be combined with `test 'a and x rows shouldn't allow non efer, solo, first parts`. That would cut this diff down by 1/3. See `user_test.rb`

test/models/user_test.rb
```diff
@@ -0,0 +1,176 @@
...
+      user = build(:admin, instrument: User.instruments[:sousaphone], part: User.parts[key])
+      refute user.valid?
+      assert_equal "Admin with instrument sousaphone can't have part #{key}", user.errors.full_messages.join
+    end
+  end
+
+  test 'two users can\'t have the same spot' do
+    user_a = create(:user)
+    user_b = build(:user, spot: user_a.spot)
+    refute user_b.valid?
```
atareshawty: `user_b` is invalid for reasons other than a non-unique spot. Fix that

PR #263, opened 2017-02-21
move some stuff around to add proxy for webpack server

atareshawty:
This is a WIP PR. In order to allows cors, we're adding a proxy from `localhost:3000` (webpack dev server) to `localhost:3001` (rails server)


client/src/utils/api.js
```diff
@@ -5,7 +5,7 @@ import errorEmitter from './error-emitter';
 
 const { getToken, refreshToken } = auth;
 const baseRoute = process.env.NODE_ENV === 'development'
-  ? `http://localhost:${process.env.SERVER_PORT}`
+  ? `http://localhost:${3001}`
```
atareshawty: We can leave this line alone now

PR #267, opened 2017-02-28
Challenge model

atareshawty:


app/models/performance.rb
```diff
@@ -1,11 +1,15 @@
...
+  has_many :challenges
+
   # validations
   validates :name, presence: true
   validates :date, presence: true, uniqueness: true
   validates :window_open, presence: true, uniqueness: true
   validates :window_close, presence: true, uniqueness: true
 
   validate :window_open_before_window_close
+  # validate :uniqueness of spot in challenge for performance ie) A4 can only be challenged once for the Oklahoma Game
```
atareshawty: You should probably do this and test for it

atareshawty:
Travis is failing (thanks amazon), but the tests pass locally so I'm merging

PR #270, opened 2017-03-02
Profile

atareshawty:
Added `user#show` to `users_controller`


client/src/utils/is_empty_object.js
```diff
@@ -0,0 +1,3 @@
+export default function isEmptyObject(obj = {}) {
```
atareshawty: This should default to an empty object. If you don't pass the function, it shouldn't return true

PR #271, opened 2017-03-02
Set RAILS_ENV globally and use db:reset

tarebyte:
@atareshawty I _think_ this should solve your travis database woes.


atareshawty:
Let's wait a minute and see 😜

atareshawty:
🎉 🍻 Thanks!

PR #272, opened 2017-03-04
React

atareshawty:
This is a working PR of a few things:

- Move from npm to yarn (hence the giant yarn.lock file)
- Begin the process of moving all material-ui components to use [google's implementation](https://github.com/material-components/material-components-web)


atareshawty:
K. So a few things are happening here. In the spirit of maintainability, the react code is following a new convention. ([requiredreading](https://medium.com/@alexmngn/how-to-better-organize-your-react-applications-2fd3ea1920f1#.cdzv8kq1r)).

This app is no longer supporting Safari's 'home screen web app thing'. There's a significant performance decrease. I can't find anything more relevant than 2011 on the matter, so it's not worth it.

client/src/components/drawer/index.jsx
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  handleDrawerClick(e) {
+    if (this.props.onClick) {
+      this.props.onClick(e);
+    }
+  }
+
+  handleEscape(e) {
+    if (e.code === 'Escape' && this.props.onCloseRequest) {
```
atareshawty: Check to see if this is the same across all browsers

atareshawty: It's supported in Chrome, Firefox, and (most importantly) Safari. [link](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code)

PR #275, opened 2017-03-31
Create performance

atareshawty:
Along with performance creation, we're making every `view.json.jbuilder` in snake_case and converting to camelCase in the client. The `changeCase` is a util and is used in api.js


atareshawty:
This works on staging, so :shipit:

PR #280, opened 2017-05-05
WIP Switch Testing Framework to RSpec

kylekthompson:
\cc @atareshawty


spec/requests/users_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+    }
+  end
+
+  describe 'POST /api/users/:id/reset_password/' do
+    it 'successfully resets a password' do
+      post endpoint, params: reset_params.to_json, headers: unauthenticated_header
+
+      expect(response).to have_http_status(204)
+      expect(BCrypt::Password.new(user.reload.password_digest)).to eq(new_password)
+      expect(prr.reload.used).to be(true)
```
atareshawty: Does rspec not have anything like `assert`?

spec/requests/users_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+    end
+
+    context 'when the reset request is used' do
+      let(:prr) { create(:used_password_reset_request, user: user) }
+
+      it 'returns a 403' do
+        expect {
+          post endpoint, params: reset_params.to_json, headers: unauthenticated_header
+        }.not_to change { user.reload }
```
atareshawty: Alright, I really like this

spec/requests/users_spec.rb
kylekthompson: Per @tarebyte you can use `assert` in here ¯\_(ツ)_/¯

kylekthompson:
> One question: Why the change from test(spec)/controllers/ to test(spec)/requests? I noticed that both the old controller tests went in requests along with the integration tests.

In RSpec, [controller tests](https://www.relishapp.com/rspec/rspec-rails/docs/controller-specs) stub out views and are meant to simulate a singe request and its outcomes. Currently the way you have your tests written are more like their [request specs](https://www.relishapp.com/rspec/rspec-rails/v/3-6/docs/request-specs/request-spec) which will actually render your views (and take your urls) and behave as a sort of "integration" test for requests.

spec/requests/users_spec.rb
tarebyte: The proper RSpec syntax would be

```ruby
expect(prr.reload.used).to be_truthy
```

kylekthompson: `be_truthy` matches `true`, `1`, `'some string'`, etc. I'm looking specifically for `true`

kylekthompson:
Codecov no 😢

spec/models/challenge_spec.rb
```diff
@@ -0,0 +1,147 @@
...
+          .to include('Users must be unique in a challenge')
+      end
+    end
+  end
+
+  context 'open spot challenge' do
+    subject { build(:open_spot_challenge) }
+
+    it 'is invalid with no users' do
+      skip('how is this supposed to work? currently fails')
```
kylekthompson: I think this is why codecov went down a tiny bit.

atareshawty: I missed the original problem with Codecov. What's wrong?

kylekthompson: > Merging #280 into rails will decrease coverage by 0.66%.

atareshawty: The validation is wrong. The test is good. This:

```ruby
  def valid_open_spot_challenge
    return if normal_challenge_type? || tri_challenge_type?
    return if users.length <= 2
    errors.add(:users, 'no more than two users are allowed in an open spot challenge')
  end
```

Should look something like this instead:

```ruby
  def valid_open_spot_challenge
    return unless open_spot_challenge_type?
    return if !users.empty? && users.length <= 2
    errors.add(:users, 'no more than two users are allowed in an open spot challenge')
  end
```

kylekthompson: 👍 I'll fix that.

spec/models/discipline_action_spec.rb
```diff
@@ -0,0 +1,6 @@
+require 'rails_helper'
+
+describe DisciplineAction, type: :model do
+  it { is_expected.to validate_presence_of(:reason) }
```
atareshawty: These are really cool

spec/models/performance_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  it { is_expected.to validate_uniqueness_of(:window_close) }
+
+  describe 'validations' do
+    let(:performance) { build(:performance, window_open: window_open, window_close: window_close) }
+
+    context 'when window_open is earlier than window_close' do
+      let(:window_open) { Time.zone.now }
+      let(:window_close) { Time.zone.now + 1.day }
+
+      specify { expect(performance).to be_valid }
```
atareshawty: Can you explain this block? Are you using `specify` rather than `it`?

kylekthompson: Correct. `specify` is just a shorter syntax that you can use when an `it` block would be kinda useless.

In this case, the `it` block would just be something like:

```ruby
it 'is valid' do
  expect(performance).to be_valid
end
```

The 'is valid' is redundant when there is already the `be_valid` there

Some people prefer the `it`-only way, though, so I can change it if you'd like 👍

spec/models/user_challenge_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    context 'when the challenge is normal' do
+      let(:challenge) { build(:normal_challenge, spot: challengee.spot, performance: performance) }
+
+      before do
+        challenge.users = [challengee, challenger]
+      end
+
+      it 'lets the challenger leave the challenge' do
+        expect(
+          described_class.can_user_remove_self_from_challenge?(challenger, challenge, challenge.user_challenges[1])
```
atareshawty: In other test files, you used something called `subject`. In this, you're using a `before` block and `described_class`. Can you explain why?

spec/models/performance_spec.rb
kylekthompson: RSpec also cleverly englishifies the words for this:

> Performance validations when window_open is earlier than window_close should be valid

spec/models/user_challenge_spec.rb
kylekthompson: `subject` is an instance of the described class, typically. `described_class` is the class itself.

If this were an instance method, I would be using `subject` (and still using a `before` block because adding users to the challenge is setup required for what we are testing here).

kylekthompson: Nothing is stopping you from doing something like this, though:

```ruby
subject { described_class }
```

spec/models/performance_spec.rb
atareshawty: > so I can change it if you'd like

No change necessary. Thank you for the explanation! I agree

spec/models/user_challenge_spec.rb
atareshawty: Thanks!!

spec/models/challenge_spec.rb
```diff
@@ -0,0 +1,146 @@
...
+      let(:user) { build(:user) }
+
+      it 'is full with two users' do
+        subject.users = [user, user]
+
+        expect(subject.full?).to be(true)
+      end
+
+      it 'is not full with less than two users' do
+        expect(subject.full?).to be(false)
```
kylekthompson: What do you prefer here?

This:

```ruby
expect(subject.full?).to be(false)
```

or 

```ruby 
expect(subject).not_to be_full
```

Both are valid RSpec (they turn `be_something` into `thing.something?` calls).

I feel like I prefer how it is now, because it explicitly shows what is being tested, but I could really go either way on it.

kylekthompson: If you use the `be_full` syntax you can also use specify here instead.

atareshawty: I'm fine with how it is now

PR #281, opened 2017-05-11
Roster

atareshawty:
WIP PR for user roster table with spot and part editing

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,8 +1,10 @@
 class UsersController < ApplicationController
   before_action :authenticate_user, except: [:reset_password]
-  before_action :ensure_admin!, except: [:profile, :reset_password]
-  before_action :ensure_correct_user!, only: [:profile]
+  before_action :ensure_admin!, except: [:profile, :reset_password, :update]
+  before_action :ensure_correct_user!, only: [:profile, :update]
   before_action :ensure_password_reset_request_is_valid!, only: [:reset_password]
+  before_action :ensure_new_part_exists!, only: [:update]
+  before_action :ensure_target_spot_exists!, only: [:switch_spot]
 
   def index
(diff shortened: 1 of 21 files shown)


spec/requests/users_spec.rb
```diff
@@ -71,3 +71,21 @@
...
+    {
+      part: 'first'
+    }
+  end
+  describe 'PUT /api/users/:id' do
+    it 'successfully updates a user\'s part' do
+      put endpoint, params: update_params.to_json, headers: authenticated_header(user)
+
+      expect(response).to have_http_status(200)
+      user.part.should eq(User.parts[update_params['part']])
```
kylekthompson: Just a heads up, `should` shouldn't be used anymore. It was replaced [in 2012](http://rspec.info/blog/2012/06/rspecs-new-expectation-syntax/) because of some issues (mentioned in the article I linked).

I think their DNS is down right now, but http://betterspecs.org/ is a good resource for RSpec best practices.

Also, you can consider adding [rubocop-rspec](https://github.com/backus/rubocop-rspec) for spec linting if you want (it'll catch stuff like this)

atareshawty: Thanks for the heads up. That's been fixed locally. I just haven't pushed in a while. RSpec has nice deprecations when you run the tests. Thanks for the resources!

app/controllers/users_controller.rb
```diff
@@ -1,22 +1,58 @@
...
+    @users = User.performers.includes(:spot).sort { |a, b| a.spot <=> b.spot }
   end
+  # rubocop:enable Performance/CompareWithBlock
 
   def show
     @user = User.includes(:spot, :challenges, :discipline_actions).find(params[:buck_id])
   end
 
+  def update
+    logger.info 'HEY'
```
atareshawty: Get rid of this

PR #283, opened 2017-05-12
User

atareshawty:
It was a bad idea to have an `id` and a `buck_id` with which to identify a user. This moves everything to just use `buck_id`. 

While I was at it, the migrations were condensed. Since there isn't a production database yet, combining some migrations makes sense


app/models/user.rb
```diff
@@ -37,13 +37,13 @@ class User < ApplicationRecord
   validate :valid_instrument_part_for_user
   validate :valid_instrument_part_for_admin
 
-  before_validation :downcase_buck_id
+  # before_validation :downcase_buck_id
```
atareshawty: lol just get rid of this line

PR #287, opened 2017-06-03
Rails: add ChallengeBylder

atareshawty:
Attempting to use the builder pattern to take a lot of logic out of the challenge controller.

And adding the fact that there are rows that can't participate in tri challenges


app/models/challenge/bylder.rb
```diff
@@ -0,0 +1,50 @@
+# Builder to localize challenge creation logic
+class Challenge
+  class Bylder
```
tarebyte: 😆

PR #291, opened 2017-06-04
Set local Ruby version

tarebyte:
Ran into this yesterday

```
❯ bundle install
Your Ruby version is 2.4.1, but your Gemfile specified 2.4.0
```

This fixes that issue by running `rbenv local 2.4.0` and setting a local version.


atareshawty:
Thanks!!

PR #303, opened 2017-07-25
Challenge Evaluations API

kylekthompson:
@atareshawty let me know what you think about the direction of this so far

Also, how much do you want this broken up? It's already getting big and this isn't even in the API yet.

### DIFF
--- a/app/controllers/user_challenges_controller.rb
+++ b/app/controllers/user_challenges_controller.rb
@@ -35,7 +35,10 @@ class UserChallengesController < ApplicationController
     if result.success?
       head :no_content
     else
-      render json: { resource: 'user_challenge', errors: [user_challenge: result.errors.first] }, status: 500
+      render json: {
+        resource: 'user_challenge',
+        errors: [user_challenge: result.errors.first]
+      }, status: :unprocessable_entity
     end
   end
 
@@ -46,7 +49,10 @@ class UserChallengesController < ApplicationController
     if result.success?
       head :no_content
     else
-      render json: { resource: 'user_challenge', errors: [user_challenge: result.errors.first] }, status: 500
+      render json: {
+        resource: 'user_challenge',
+        errors: [user_challenge: result.errors.first]
+      }, status: :unprocessable_entity
     end
   end
 
--- a/app/controllers/user_challenges_controller.rb
+++ b/app/controllers/user_challenges_controller.rb
@@ -4,6 +4,7 @@ class UserChallengesController < ApplicationController
   before_action :ensure_user_exists!, only: [:create]
   before_action :ensure_challenge_not_full!, only: [:create]
   before_action :ensure_correct_user_for_delete!, only: [:destroy]
+  before_action :ensure_user_can_evaluate, only: [:evaluate_comments, :evaluate_places]
 
   def create
     challenge = Challenge.find_by id: params[:challenge_id]
@@ -24,6 +25,28 @@ class UserChallengesController < ApplicationController
       remove_user_from_open_spot_challenge uc
     else
       remove_user_from_non_open_spot_challenge uc
+    end
+  end
+
+  def evaluate_comments
+    evaluator = UserChallenge::Evaluator.new(params: params)
+    result = evaluator.save_comments
+
+    if result.success?
+      head :no_content
+    else
+      render json: { resource: 'user_challenge', errors: [user_challenge: result.errors.first] }, status: 500
+    end
+  end
+
+  def evaluate_places
+    evaluator = UserChallenge::Evaluator.new(params: params)
+    result = evaluator.save_places
+
+    if result.success?
+      head :no_content
+    else
+      render json: { resource: 'user_challenge', errors: [user_challenge: result.errors.first] }, status: 500
     end
   end
 
@@ -72,5 +95,14 @@ class UserChallengesController < ApplicationController
     challenge = user_challenge.challenge
     return if UserChallenge.can_user_remove_self_from_challenge? user, challenge, user_challenge
     render json: { resource: 'user_challenge', errors: [user: 'doesn\'t have access to that challenge'] }, status: 403
+  end
+
+  def ensure_user_can_evaluate
+    challenge = UserChallenge.find(params[:user_challenges].first.try(:[], :id)).challenge
+    return if challenge.can_be_evaluated_by?(user: current_user)
+    render json: {
+      resource: 'user_challenge',
+      errors: [user: 'doesn\'t have permission to update that challenge']
+    }, status: :unauthorized
   end
 end
--- a/app/controllers/user_challenges_controller.rb
+++ b/app/controllers/user_challenges_controller.rb
@@ -4,7 +4,7 @@ class UserChallengesController < ApplicationController
   before_action :ensure_user_exists!, only: [:create]
   before_action :ensure_challenge_not_full!, only: [:create]
   before_action :ensure_correct_user_for_delete!, only: [:destroy]
-  before_action :ensure_user_can_evaluate, only: [:evaluate_comments, :evaluate_places]
+  before_action :ensure_user_can_evaluate!, only: [:evaluate_comments, :evaluate_places]
 
   def create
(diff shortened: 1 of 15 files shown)


app/models/user_challenge/evaluator.rb
```diff
@@ -0,0 +1,43 @@
...
+      save_a_field(field_name: :comments, id_map: ids_to_comments)
+    end
+
+    def save_places
+      save_a_field(field_name: :place, id_map: ids_to_places)
+    end
+
+    private
+
+    def save_a_field(field_name:, id_map:)
```
atareshawty: Can you explain this syntax? (field_name:)

app/models/result.rb
```diff
@@ -0,0 +1,25 @@
+class Result
```
kylekthompson: It's useful to track whether pretty much anything was successful or not! I'm just using it here for that reason.

It lets you return the truthy/falsey "was this a success?" while also giving back either errors or the result of whatever you just did.

(also shamelessly stole this from @tarebyte because I loved the idea of it when I saw it in Classroom)

app/models/user_challenge/evaluator.rb
kylekthompson: Yup!

`def method_name(arg_name:)`:
- In this case, `arg_name` is a required keyword argument. Meaning if I don't call the method like `method_name(arg_name: some_arg_here)`, it'll throw an `ArgumentError`.

`def method_name(arg_name: nil)`:
- In this case, `arg_name` is a keyword argument with a default value (of `nil`), so it is optional.

[Here's](https://robots.thoughtbot.com/ruby-2-keyword-arguments) a nice blog post on them.

kylekthompson:
> I like this so far! I'm cool if this gets a little bigger since I've looked over it so far. I left a few comments, but here's a few more.

✨ 🎉 🎊  

> I think we talked about it, but can place be an enum on UserChallenge? I realize I originally said just a number, but I now realize I was wrong.

Will do!

> I think the scope of this PR can be broadened to actually add the API end points/tests. Definitely no front end code though.

👍

spec/models/result.rb
```diff
@@ -0,0 +1,46 @@
+require 'rails_helper'
+
+describe Result, type: :model do
+  let(:a_result) { 'a result' }
+  let(:errors) { ['error'] }
+
```
kylekthompson: Note to self: add tests for `#success?` and `#failure?` here

kylekthompson:
@atareshawty So I'm not quite sure when it makes sense to transition to `:needs_approval` in here.

And depending on when that happens, we need to think about validating the places on user challenges (to ensure there is the correct combo of first, second, and maybe third).

My initial thoughts were something like this:

1. Save the comments
2. Save the places. At this point we can't do validations to ensure the places are correct because they might change their mind (and they we would have overlapping place assignments until they are all updated)
3. Submit for approval. At this point, we can run validations to ensure the places have been properly assigned, because they won't be changing anymore

app/models/challenge.rb
```diff
@@ -36,6 +35,16 @@ def self.tri_challenge_rows
     [:j]
   end
 
+  # rubocop:disable Metrics/PerceivedComplexity
+  def can_be_evaluated_by?(user:)
```
kylekthompson: Please please please double and triple check this (along with the tests)! 😅

atareshawty:
I think the UI/UX should drive the api decision here. How about you can update the comments/rankings as you please until you submit it for approval? Then the validations would happen.

Everything can update individually while a challenge is before the `:needs_approval` stage. When the challenge gets moved to `:needs_approval`, it then checks if that's possible (the rankings are valid).

Basically what you said, I'm just making sure we're on the same page 😝 

In terms of UI flow, a user would be able to go back and forth from saving comments and rankings. The updating wouldn't have to be in any particular order.

app/controllers/user_challenges_controller.rb
```diff
@@ -4,6 +4,7 @@ class UserChallengesController < ApplicationController
   before_action :ensure_user_exists!, only: [:create]
   before_action :ensure_challenge_not_full!, only: [:create]
   before_action :ensure_correct_user_for_delete!, only: [:destroy]
+  before_action :ensure_user_can_evaluate, only: [:evaluate_comments, :evaluate_places]
```
atareshawty: Slight nit, other before_actions end in a `!`. I'm not sure what Rails convention dictates, but that's what I've been doing

spec/requests/user_challenge_evaluation_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+
+describe 'User Challenges Evaluation', type: :request do
+  subject(:request) { post endpoint, params: params.to_json, headers: authenticated_header(admin) }
+  let(:admin) { create(:admin_user) }
+  let(:challenge) { create(:normal_challenge) }
+  let(:params) { { user_challenges: user_challenge_params } }
+
+  describe 'POST /api/user_challenges/comments' do
+    let(:endpoint) { '/api/user_challenges/comments' }
+    let(:comments_1) { 'User 1 did some things.' }
```
atareshawty: Very creative comments

kylekthompson:
> I think the UI/UX should drive the api decision here. How about you can update the comments/rankings as you please until you submit it for approval? Then the validations would happen.

Yep! I think we're on the same page here

> Everything can update individually while a challenge is before the :needs_approval stage. When the challenge gets moved to :needs_approval, it then checks if that's possible (the rankings are valid).

Still agree here 👍 

> Basically what you said, I'm just making sure we're on the same page 😝

✨ 

> In terms of UI flow, a user would be able to go back and forth from saving comments and rankings. The updating wouldn't have to be in any particular order.

Yep!

---

Cool. Just wanted to make sure I wasn't missing anything important on the validation front. I'll probably handle that once I do the UI if you're cool with that? Seems like it pairs up with that part better than this part.

atareshawty:
> I'll probably handle that once I do the UI if you're cool with that

Fine with me! Since the api behavior depends on the UI it makes sense to organize that way

app/controllers/user_challenges_controller.rb
kylekthompson: Will do!

I usually only `!` when something `raise`s, but I'll update that!

spec/requests/user_challenge_evaluation_spec.rb
kylekthompson: I try ¯\\\_(ツ)\_/¯

app/controllers/user_challenges_controller.rb
```diff
@@ -27,6 +28,28 @@ def destroy
...
   end
 
+  def evaluate_comments
+    evaluator = UserChallenge::Evaluator.new(params: params)
+    result = evaluator.save_comments
+
+    if result.success?
+      head :no_content
+    else
+      render json: { resource: 'user_challenge', errors: [user_challenge: result.errors.first] }, status: 500
```
tarebyte: Drive by comment, a [`500`](https://httpstatuses.com/500) status is probably what you want to go for here.

500's should only be raised when something like an exception is raised or something on the server is really borked.

I would look at something in the `4xx` range.

tarebyte: On GitHub we raise a [`422`](https://httpstatuses.com/422) for something like this

kylekthompson: Yea a 422 is usually what I'll do in my projects, but the rest of the actions were 500s, so I was following it for consistency.

@atareshawty if you want I can run through change status codes!

atareshawty: Sure! Thanks :D

kylekthompson: 👋 Hi @atareshawty!

So I'm just going to swap out the status codes for my actions for this PR.

I saw a few other things ([needz moar transaction](https://github.com/osumb/challenges/blob/rails/app/controllers/user_challenges_controller.rb#L34)) while I was going to do that to the other 500s, and to avoid bloating this PR I'm just going to open a new PR and do some boy scouting.

PR #304, opened 2017-07-28
Boyscouting

kylekthompson:
~~**Note: this is waiting for #303 to be merged. After that I'll rebase this!**~~

### DIFF
--- a/app/controllers/user_challenges_controller.rb
+++ b/app/controllers/user_challenges_controller.rb
@@ -62,12 +62,12 @@ class UserChallengesController < ApplicationController
   private
 
   def ensure_challenge_exists!
-    return unless Challenge.find_by(id: params[:challenge_id]).nil?
+    return if Challenge.exists? id: params[:challenge_id]
     render json: { resource: 'user_challenge', errors: [challenge: 'challenge doesn\'t exist'] }, status: 403
   end
 
   def ensure_user_exists!
-    return unless User.find_by buck_id: params[:buck_id].nil?
+    return if User.exists? buck_id: params[:challenger_buck_id]
     render json: { resource: 'user_challenge', errors: [user: 'user doesn\'t exist'] }, status: 403
   end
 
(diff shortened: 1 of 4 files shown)


app/controllers/user_challenges_controller.rb
```diff
@@ -17,45 +18,56 @@ def create
...
+
   def ensure_challenge_exists!
-    return unless Challenge.find_by(id: params[:challenge_id]).nil?
+    return if Challenge.exists? id: params[:challenge_id]
     render json: { resource: 'user_challenge', errors: [challenge: 'challenge doesn\'t exist'] }, status: 403
   end
 
   def ensure_user_exists!
-    return unless User.find_by buck_id: params[:buck_id].nil?
+    return if User.exists? buck_id: params[:challenger_buck_id]
```
kylekthompson: This was originally doing two things wrong:
```ruby
return unless User.find_by(buck_id: params[:buck_id].nil?)
```

`params[:challenger_buck_id]` is what you are actually looking for, but you were getting false positives because of the parenthesis issue

app/controllers/user_challenges_controller.rb
```diff
@@ -17,45 +18,56 @@ def create
...
+        errors: [user_challenge: result.errors.first]
+      }, status: :unprocessable_entity
     end
   end
 
+  private
+
   def ensure_challenge_exists!
-    return unless Challenge.find_by(id: params[:challenge_id]).nil?
+    return if Challenge.exists? id: params[:challenge_id]
```
kylekthompson: `exists?` is more performant because it doesn't instantiate any objects

kylekthompson:
Another general comment (that I can't do inline because the files aren't in the diff):

I wanted to go through and take a look at anywhere you are doing a `select` or `reject`. Most of those should probably be `where`s to let the database handle that kind of filtering. Although it also might not be worth the effort because your largest array wouldn't likely be anything more than a few hundred right?

atareshawty:
> Although it also might not be worth the effort because your largest array wouldn't likely be anything more than a few hundred right?

We can look at performance stuff later. Right now, the most important thing is to have all of the pieces working. Though on such a small data set, it might not make a noticeable difference.

atareshawty:
Wow! I love all of these helpers. They really clean up the controller methods. Thank you :D

PR #306, opened 2017-08-02
Challenge Evaluations API Additions

kylekthompson:
This adds some additional endpoints that I need for the front end and fixes a nasty authorization bug (I'll point out the line)

### DIFF
--- a/app/controllers/challenges_controller.rb
+++ b/app/controllers/challenges_controller.rb
@@ -19,6 +19,16 @@ class ChallengesController < ApplicationController
     end
   end
 
+  def for_evaluation
+    @evaluable_challenges = Challenge.evaluable(current_user)
+
+    if @evaluable_challenges.any?
+      render :for_evaluation, status: :ok
+    else
+      render json: { resource: 'challenge', errors: [challenge: 'not found'] }, status: :not_found
+    end
+  end
+
   private
 
   def ensure_not_challenging_alternate!
--- a/app/controllers/challenges_controller.rb
+++ b/app/controllers/challenges_controller.rb
@@ -23,11 +23,7 @@ class ChallengesController < ApplicationController
   def for_evaluation
     @evaluable_challenges = Challenge.evaluable(current_user)
 
-    if @evaluable_challenges.any?
-      render :for_evaluation, status: :ok
-    else
-      render json: { resource: 'challenge', errors: [challenge: 'not found'] }, status: :not_found
-    end
+    render :for_evaluation, status: :ok
   end
 
   def submit_for_approval
--- a/app/controllers/challenges_controller.rb
+++ b/app/controllers/challenges_controller.rb
@@ -2,6 +2,7 @@ class ChallengesController < ApplicationController
   before_action :authenticate_user
   before_action :ensure_not_challenging_alternate!, only: [:create]
   before_action :ensure_performance_is_current_and_open!, only: [:create]
+  before_action :ensure_user_can_submit_for_approval!, only: [:submit_for_approval]
   before_action :ensure_user_has_not_already_made_challenge!, only: [:create]
   before_action :ensure_user_is_challenging_correct_instrument_and_part!, only: [:create]
   before_action :ensure_spot_has_not_been_challenged!, only: [:create]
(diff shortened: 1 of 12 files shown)


app/controllers/user_challenges_controller.rb
```diff
@@ -104,8 +104,8 @@ def ensure_correct_user_for_delete!
   end
 
   def ensure_user_can_evaluate!
-    challenge = UserChallenge.find(params[:user_challenges].first.try(:[], :id)).challenge
-    return if challenge.can_be_evaluated_by?(user: current_user)
+    evaluable_user_challenge_ids = Challenge.evaluable(current_user).flat_map { |c| c.user_challenges.ids }
+    return if params[:user_challenges].all? { |uc| evaluable_user_challenge_ids.include?(uc[:id]) }
```
kylekthompson: This is where the nasty auth bug was fixed!

app/controllers/challenges_controller.rb
```diff
@@ -19,6 +20,26 @@ def create
     end
   end
 
+  def for_evaluation
+    @evaluable_challenges = Challenge.evaluable(current_user)
```
atareshawty: Evaluable is a weird word

app/controllers/user_challenges_controller.rb
atareshawty: 🎉

app/models/challenge.rb
```diff
@@ -21,6 +21,21 @@ class Challenge < ApplicationRecord
   validate :no_duplicate_challenged_spots
   validate :correct_row_for_challenge_type
 
+  # scopes
+  scope :with_users_and_spots, -> { includes(user_challenges: { user: :spot }) }
+  scope :needs_comments, -> { where(stage: :needs_comments) }
+  scope :evaluable, lambda { |user|
```
atareshawty: This file's diff is wonderful

app/controllers/challenges_controller.rb
kylekthompson: I agree though I can't think of a better one.. The internet says evaluatable is [not even close to evaluable](https://english.stackexchange.com/questions/304948/evaluable-vs-evaluatable) in usage. So probably not actually a word

kylekthompson:
@atareshawty I think this is all I should need, though we can definitely leave it open 'til the front end is done (hopefully tonight, maybe tomorrow night)

app/controllers/challenges_controller.rb
```diff
@@ -19,6 +20,26 @@ def create
...
+
+    if @evaluable_challenges.any?
+      render :for_evaluation, status: :ok
+    else
+      render json: { resource: 'challenge', errors: [challenge: 'not found'] }, status: :not_found
```
kylekthompson: I feel like `:not_found` might not be right here. Maybe I should just return the empty collection? Otherwise the UI is going to crap out instead of just saying "nothing to evaluate"

atareshawty: Yeah, since you brought it up, sending this as an error doesn't feel right. Having no results doesn't mean an error occurred. I feel like you can get rid of the check altogether. You can show custom error messages with the api util in the UI, but that would give the impression that the user did something wrong when that's not the case

kylekthompson: Yea when I wrote it I was thinking "well if nothing is returned it means they are auth'd to see them" instead of "maybe there just aren't any to evaluate"

spec/support/shared_examples/challenge.rb
```diff
@@ -0,0 +1,75 @@
+shared_examples_for 'a valid transition to :needs_approval' do |challenge_factory|
```
kylekthompson: @atareshawty mind verifying that this is correct? I might have missed something. 

Are comments mandatory?

atareshawty: Having this restriction is good. Comments aren't mandatory

app/models/user_challenge.rb
```diff
@@ -1,6 +1,6 @@
 class UserChallenge < ApplicationRecord
   # enums
-  enum place: { first: 1, second: 2, third: 3 }, _suffix: true
+  enum place: { first: 1, second: 2, third: 3 }, _prefix: true
```
atareshawty: Good catch. `first_place` is much better than `place_first`

kylekthompson:
@atareshawty should be good to go!

Issue #312, opened 2017-08-17
Previous Challenge Editing


atareshawty:
Need:
- Squad leaders have requested the ability to update comments for challenges that they've already submitted for approval
- The UI should be very similar to the one needed in #309 so there should be a lot of reusable components there
- An api endpoint that allows this

PR #313, opened 2017-08-18
Challenge comment update api

kylekthompson:
@atareshawty this should be ready for #312.

Heavily based on #303 (and reuses/refactors some of that code)


atareshawty:
LGTM!

PR #315, opened 2017-08-22
Challenge approval

### LINKED ISSUE #309: Challenge Approval
Need:
- UI scene for approving challenges after they've been evaluated and moved to `:needs_approval`
- API endpoints to for the UI to use
- Once a challenge is approved, spots should be switched based on the results (there are rules for that in `./docs/`

Details:
- Only admins should be able to access the challenge approval UI and have access to the endpoint
- Only when challenges are moved to `:done` can members involved see them
- In the same UI, admins should be able to edit the results of challenges as they see fit (or move them back to the first stage)
- There is an opportunity to create a few reusable components here for challenge result viewing in the UI

atareshawty:
WIP PR

- Api endpoint to approve a challenge
- Spot Switching Logic once last challenge for a performance is approved
  - Associated email

Need:

- Simple UI for approving challenges

Fixes #309


app/controllers/challenges_controller.rb
```diff
@@ -42,6 +43,27 @@ def submit_for_approval
...
 
+  # rubocop:disable Metrics/MethodLength
+  def approve
+    challenge = Challenge.find(params[:id])
+
+    if challenge.update(stage: :done)
+      if challenge.performance.challenges.all?(&:done_stage?)
+        switcher = Challenge::SpotSwitcher.new(challenge.performance)
+        begin
+          switcher.run!
```
kylekthompson: This looks (from a brief glance) like it might slow down the request a lot w/ all of those database hits. Maybe consider looking into something like [sidekiq](https://github.com/mperham/sidekiq) to handle this in the background. They have some [docs](https://github.com/mperham/sidekiq/wiki/Deployment#heroku) on getting things set up with Heroku.

atareshawty: That's the plan. I'm just gonna do this at a later time

PR #338, opened 2017-09-03
Downcase buck_id when we get it from params

### LINKED ISSUE #336: Downcase Lots of Places
Where anything related to auth is involved, `.downcase` the buck_id or user email getting passed from the client

kylekthompson:
Fixes #336 

I just did a regex search for `params.*buck_id`. I'll do another pass to make sure I got all of them though.

### DIFF
--- a/spec/requests/password_reset_requests_spec.rb
+++ b/spec/requests/password_reset_requests_spec.rb
@@ -32,5 +32,17 @@ describe 'Password Reset Requests', type: :request do
         expect(response).to have_http_status(403)
       end
     end
+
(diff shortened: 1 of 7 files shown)


spec/requests/password_reset_requests_spec.rb
```diff
@@ -32,5 +32,17 @@
         expect(response).to have_http_status(403)
       end
     end
+
+    context 'when an uppercase buck_id is passed' do
```
atareshawty: Nice test! Thanks

PR #344, opened 2017-09-03
Hide Challenge

### LINKED ISSUE #343: User shouldn't see that they're challenging their own spot
It currently shows in a user's profile that they're getting challenged

atareshawty:
Fixes #343

### DIFF
--- a/app/views/users/profile.json.jbuilder
+++ b/app/views/users/profile.json.jbuilder
@@ -11,7 +11,9 @@ else
     json.partial! 'performances/performance', performance: @next_performance
   end
 
-  if !@user.challenges.last.nil? && @user.challenges.last.performance.id == @next_performance.id
+  if @user.challenges.length <= 0 || @user.challenges.last.spot == @user.spot
+    json.current_challenge nil
+  elsif !@user.challenges.last.nil? && @user.challenges.last.performance.id == @next_performance.id
     json.current_challenge do
       json.challenge_type @user.challenges.last.challenge_type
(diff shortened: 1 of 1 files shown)


app/views/users/profile.json.jbuilder
```diff
@@ -11,7 +11,9 @@ else
     json.partial! 'performances/performance', performance: @next_performance
   end
 
-  if !@user.challenges.last.nil? && @user.challenges.last.performance.id == @next_performance.id
+  if @user.challenges.length <= 0 || @user.challenges.last.spot == @user.spot
```
kylekthompson: You should swap out `length` for `count` here so it uses a `COUNT` query instead of letting ruby check the length of the challenges.

atareshawty: It's going to make the query later anyway, right? IMO, it's not that big of a deal

kylekthompson: Nope. It's grabbing the last one later. This query makes active record load all of the challenges into memory.

atareshawty: All < 10 of them? 😝

kylekthompson: Unless you're passing all of the challenges somewhere else in this file (which I didn't look for)

atareshawty: > Unless you're passing all of the challenges somewhere else in this file (which I didn't look for)

Yeah. Later on, all the challenges get rendered anyway

PR #345, opened 2017-09-04
Fix Tri Challenge

### LINKED ISSUE #341: Broken J row challenges
`ChallengeBuilder#associated_user_for_tri_challenge` is where the bug is. The code assumes that only 3 performers of each J row part are actually in J row. In reality, it's more like 5 or 6

atareshawty:
Fixes #341

### DIFF
--- a/app/models/challenge/bylder.rb
+++ b/app/models/challenge/bylder.rb
@@ -28,7 +28,7 @@ class Challenge
         if type == Challenge.challenge_types[:normal]
           users << associated_user_for_normal_challenge(spot)
         elsif type == Challenge.challenge_types[:tri]
-          users << associated_user_for_tri_challenge(challenger)
+          users << associated_users_for_tri_challenge(challenger, spot)
         end
         users.flatten
       end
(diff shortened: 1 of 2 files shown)


app/models/challenge/bylder.rb
```diff
@@ -37,13 +37,15 @@ def associated_user_for_normal_challenge(spot)
         spot.user
       end
 
-      def associated_user_for_tri_challenge(challenger)
-        User.includes(:spot).where(
-          'instrument = ? and part = ? and buck_id != ?',
+      def associated_users_for_tri_challenge(challenger, spot)
+        regular_user = spot.user
+        other_alternate = User.joins('LEFT OUTER JOIN spots on spots.id = users.spot_id').find_by(
```
kylekthompson: If you want a little less SQL you can do something like this:

```ruby
User
  .left_outer_joins(:spot)
  .where(instrument: challenger.instrument, part: challenger.part)
  .where('file > ?', 12)
  .first
```

atareshawty: But I ❤️ sql

PR #352, opened 2017-09-05
Fix joining open spot challenge bug

### LINKED ISSUE #349: Bug in a second user joining an open spot challenge
Two challenges were made on an open spot instead of the second person joining the other challenge. Look at `Challenge::ChallengeBuilder`

atareshawty:
Fixes #349

### DIFF
--- a/app/models/challenge/bylder.rb
+++ b/app/models/challenge/bylder.rb
@@ -12,7 +12,7 @@ class Challenge
       private
 
       def create_or_update(performance:, spot:, challenge_type:, users:)
-        challenge = find_challenge(performance: performance, spot: spot)
+        challenge = Challenge.find_by(performance: performance, spot: spot)
         if challenge.nil?
           create(performance: performance, spot: spot, challenge_type: challenge_type, users: users)
         else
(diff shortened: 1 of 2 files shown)


app/models/challenge/bylder.rb
```diff
@@ -3,16 +3,36 @@ class Challenge
...
+        if challenge.nil?
+          create(performance: performance, spot: spot, challenge_type: challenge_type, users: users)
+        else
+          update(challenge: challenge, users: users)
+        end
+      end
+
+      def create(performance:, spot:, challenge_type:, users:)
+        challenge = Challenge.new(performance: performance, spot: spot, challenge_type: challenge_type, users: users)
+        challenge.user_challenges.each { |uc| uc.spot = uc.user.spot }
```
kylekthompson: I think I understand better what you were trying to do here now.

If you want, you can add a `before_create` callback that assigns the spot for you, though some people prefer to be more explicit and avoid those hidden side effects (e.g. in the case where you might need to set a spot explicitly, it would get overwritten unless you either check if there is already a spot or update the user challenge after creation)

app/models/challenge/bylder.rb
```diff
@@ -45,6 +72,14 @@ def associated_user_for_tri_challenge(challenger)
           challenger.buck_id
         )
       end
+
+      def associated_user_for_open_spot_challenge(spot:, performance:)
+        (find_challenge(spot: spot, performance: performance)&.users || []).first
+      end
+
+      def find_challenge(spot:, performance:)
```
kylekthompson: This seems unnecessary since it is only doing a `find_by`. When reading the code that uses this, I'd expect it to be doing something more

app/models/challenge/bylder.rb
atareshawty: I don't think there's a case when the spot logic would need to behave differently, but I do prefer to keep it out of a `before_create`

app/models/challenge/bylder.rb
atareshawty: I agree. Probably went a little overboard with this

PR #354, opened 2017-09-06
Fix User Challenge Permissions

### LINKED ISSUE #347: Permissions on User Challenges
In cross row challenges. Squad leaders are only able to see the user challenges in their row. They should be able to see all the users in an entire challenge if they're able to see one of them.

kylekthompson:
Fixes #347

### DIFF
--- a/app/models/challenge.rb
+++ b/app/models/challenge.rb
@@ -32,7 +32,8 @@ class Challenge < ApplicationRecord
     elsif user.director?
       with_users_and_spots.where(users: { instrument: user.instrument })
     elsif user.squad_leader?
-      with_users_and_spots.where(spots: { row: user.spot.row })
+      ids = with_users_and_spots.where(spots: { row: user.spot.row }).pluck(:id)
+      with_users_and_spots.where(id: ids)
     else
       none
     end
--- a/spec/models/challenge_spec.rb
+++ b/spec/models/challenge_spec.rb
@@ -236,8 +236,10 @@ describe Challenge, type: :model do
 
         specify { expect(described_class.viewable_by_user(user_1)).to include(challenge) }
         specify { expect(described_class.viewable_by_user(user_1).count).to eq(1) }
+        specify { expect(described_class.viewable_by_user(user_1).first.user_challenges.length).to eq(challenge.user_challenges.count) }
         specify { expect(described_class.viewable_by_user(user_2)).to include(challenge) }
         specify { expect(described_class.viewable_by_user(user_2).count).to eq(1) }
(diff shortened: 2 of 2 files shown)


app/models/challenge.rb
```diff
@@ -32,7 +32,8 @@ class Challenge < ApplicationRecord
     elsif user.director?
       with_users_and_spots.where(users: { instrument: user.instrument })
     elsif user.squad_leader?
-      with_users_and_spots.where(spots: { row: user.spot.row })
+      ids = with_users_and_spots.where(spots: { row: user.spot.row }).pluck(:id)
+      with_users_and_spots.where(id: ids)
```
kylekthompson: I can't think of a better way to do this in one query right now, but I'll think on it a tad.

It'd likely have to be without ActiveRecord and probably a sub-query

atareshawty: I'm fine with 2 queries so long as it's not N+1, we're good :)

spec/models/challenge_spec.rb
```diff
@@ -236,8 +236,10 @@
 
         specify { expect(described_class.viewable_by_user(user_1)).to include(challenge) }
         specify { expect(described_class.viewable_by_user(user_1).count).to eq(1) }
+        specify { expect(described_class.viewable_by_user(user_1).first.user_challenges.length).to eq(challenge.user_challenges.count) }
```
atareshawty: Question, does the data you set up here reflect cross row challenges? Like someone in A row challenging a spot in X row and users 1 and 2 are squad leaders in A and X row?

kylekthompson: Yes, but it kinda relies on the challenge factory not changing. The factory makes a user in row a and a user in the spot of the challenge, so I set the spot to be in x row:

```ruby
[2] pry(#<RSpec::ExampleGroups::Challenge::ViewableByUser::WhenASquadLeaderIsEvaluating::AndTheyAreInTheRowOfOneOfTheParticipants>)> challenge.user_challenges.map(&:spot)
=> [#<Spot:0x007fa823443c98 id: 129, row: "a", file: 13, created_at: Thu, 07 Sep 2017 01:31:18 UTC +00:00, updated_at: Thu, 07 Sep 2017 01:31:18 UTC +00:00>,
 #<Spot:0x007fa82337b6d0 id: 128, row: "x", file: 1, created_at: Thu, 07 Sep 2017 01:31:18 UTC +00:00, updated_at: Thu, 07 Sep 2017 01:31:18 UTC +00:00>]
```

kylekthompson: [Here is the code for the factory](https://github.com/osumb/challenges/blob/master/spec/factories/challenges.rb#L33-L34)

PR #360, opened 2017-09-09
Adds unique index on spots.row and spots.file to ensure unique spots

### LINKED ISSUE #340: Spot's Need To Be Unique
How this was exposed:

- Upon the mass user upload, spots with duplicate rows and numbers were created
- In the challenges controller, it's grabbing the `challengee` by getting the user from the spot that's passed. The duplicate spot doesn't have a user, so it's nil. 500 error happens because the logic assumes that a user will be there
- Challenge doesn't get made and bad things happen

atareshawty:
Fixes #340


db/migrate/20170908015143_add_unique_row_file_for_spot.rb
```diff
@@ -0,0 +1,9 @@
+class AddUniqueRowFileForSpot < ActiveRecord::Migration[5.0]
+  def up
+    add_index :spots, [:row, :file], unique: true, name: 'index_spots_on_row_and_file'
```
kylekthompson: ✨

PR #379, opened 2017-10-18
Send password reset emails during mass user upload

### LINKED ISSUE #362: Redo mass user creation flow
Sending a generated password has been problematic. That flow needs to change

kylekthompson:
Closes #362 

screen shot 2017-10-17 at 11 11 12 pm


kylekthompson:
I'm not really sure what else I can do regarding my diff coverage since the mailer `return`s if the env is test, but I can try to get that tested if you'd like.

PR #385, opened 2017-11-10
Background Jobs!

### LINKED ISSUE #357: Start Utilizing Jobs
Requests currently have to wait for emails to fire off so having some of that being async would be really nice

atareshawty:
Actually using job processing (Fixes #357 )


app/jobs/challenge_email_job.rb
```diff
@@ -0,0 +1,11 @@
+class ChallengeEmailJob < ApplicationJob
```
kylekthompson: You can do something like an `EmailJob` that looks like this if you want to only have one:

```ruby
class EmailJob < ApplicationJob
  queue_as :default

  def perform(mailer_klass:, email_method:, args:)
    mailer_klass.constantize.send(email_method, args.deep_symbolize_keys)
  end
end
```

Then call it like this:

```ruby
EmailJob.perform_later(
  mailer_klass: ChallengeSuccessMailer, 
  email_method: :challenge_success_email, 
  args: { ... }
)
```

atareshawty: I like this. Thanks!

PR #396, opened 2017-11-22
User creation

atareshawty:
{{ wip }}

### DIFF
--- a/app/services/user_service.rb
+++ b/app/services/user_service.rb
@@ -1,0 +1,58 @@
+module UserService
+  def self.create(params:)
+    new_user = _user_from_params(params)
+
+    if new_user.performer?
+      User.transaction do
+        old_user = User.find_by(current_spot: new_user.current_spot)
+        deactivate_user(buck_id: old_user.buck_id) unless old_user.nil?
+      end
+    end
+
+    new_user.save! if new_user.valid?
+    new_user
+  end
+
+  def self.deactivate_user(buck_id:)
+    user = User.find(buck_id)
+    user.active = false
+    user.current_spot = nil
+    user.original_spot = nil
+    user.save!
+    user
+  end
+
+  # rubocop:disable Metrics/MethodLength
+  def self._user_from_params(params)
+    permitted = _create_params(params)
+    spot = nil
+
+    unless permitted[:spot].nil?
+      spot_params = permitted[:spot]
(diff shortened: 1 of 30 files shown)


app/services/user_service.rb
```diff
@@ -0,0 +1,58 @@
+module UserService
+  def self.create(params:)
+    new_user = _user_from_params(params)
+
+    if new_user.performer?
+      User.transaction do
```
kylekthompson: All of this should probably be wrapped in a transaction, not just this part. If you don't, the `new_user.save!` could fail and leave the old user deactivated without a replacement.

app/services/user_service.rb
```diff
@@ -0,0 +1,58 @@
...
+      end
+    end
+
+    new_user.save! if new_user.valid?
+    new_user
+  end
+
+  def self.deactivate_user(buck_id:)
+    user = User.find(buck_id)
+    user.active = false
```
kylekthompson: This could be shorter if you want:

```ruby
user = User.find(buck_id)
user.update!(active: false, current_spot: nil, original_spot: nil)
user
```

app/services/user_service.rb
```diff
@@ -0,0 +1,58 @@
...
+    user.current_spot = nil
+    user.original_spot = nil
+    user.save!
+    user
+  end
+
+  # rubocop:disable Metrics/MethodLength
+  def self._user_from_params(params)
+    permitted = _create_params(params)
```
kylekthompson: I think this does what you are trying to do here while keeping within your method length limits:

```ruby
def self._user_from_params(params)
  user_attributes = _create_params(params).to_h.merge(password_digest: SecureRandom.base64)

  if user_attributes[:spot].present?
    row = user_attributes[:spot][:row]&.downcase&.to_sym
    file = user_attributes[:spot][:file]&.to_i
    spot = Spot.find_by(row: row, file: file)
    user_attributes.delete(:spot)
    user_attributes.merge(current_spot: spot, original_spot: spot)
  end

  User.new(user_attributes)
end
```

app/services/user_service.rb
```diff
@@ -0,0 +1,58 @@
+module UserService
```
kylekthompson: Maybe consider using `class << self` to get rid of the `self.`s. It also makes it easier to make the private methods actually private:

```ruby
module UserService
  class << self
    def create; end

    private

    def _user_from_params; end
  end
end
```

atareshawty:
Closing this to make two PRs, one for front end and one for backend

PR #410, opened 2018-01-13
Adds deploy script

atareshawty:


atareshawty:
I tested the branch logic by running `bin/deploy osumbchallenges deploy`, which deployed this branch to staging

PR #419, opened 2018-04-04
Replaces travis ci with circle ci

atareshawty:


docker-compose.yml
```diff
@@ -1,13 +1,13 @@
 version: '3'
 services:
   postgres:
-    image: "postgres:9.6.1"
+    image: "postgres:9.6.8"
```
atareshawty: I checked the heroku addons. This change is simply to match what's running in production

PR #423, opened 2018-04-13
Adds session authentication

atareshawty:
challengesauthdemo


app/controllers/application_controller.rb
```diff
@@ -1 +1,19 @@
...
+  end
+
+  def ensure_not_authenticated!
+    return unless current_user
+    redirect_to '/logged_in'
+  end
+
+  def current_user
+    if session[:buck_id]
+      @current_user ||= User.find_by(buck_id: session[:buck_id]) if session[:buck_id]
```
kylekthompson: Probably don't need the second `if session[:buck_id]` here

app/services/authentication_service.rb
```diff
@@ -0,0 +1,5 @@
+module AuthenticationService
+  def self.authenticate_user(params)
+    User.find_by(buck_id: params[:buck_id])&.authenticate(params[:password])
```
kylekthompson: This is susceptible to [timing attacks](https://en.wikipedia.org/wiki/Timing_attack) if you're worried about it.

app/controllers/application_controller.rb
atareshawty: good catch! this was a copy paste issue

PR #447, opened 2018-08-05
Fix favicon and email images

atareshawty:
[Trello](https://trello.com/c/FhbG3Acl/15-fix-favicon)


atareshawty:
Checked both of these fixes in staging and it works 🚀

